PR #545, opened 2019-04-18
Support choosing partition by schema or table

sunxiaoguang:
Support choosing partition by schema or table if users explicitly say so
otherwise use fixed partition 0.

Signed-off-by: Xiaoguang Sun <sunxiaoguang@zhihu.com>

Make drainer possible to utilize multiple kafka partitions when the global ordering of binlog events isn't  
required.

Code changes
Support choosing kafka partition by schema or table.

no
no
no
 - Has persistent data change
no

Related changes
no
yes
yes
no


sunxiaoguang:
Looks like integration test failed in a strange way.

drainer/sync/kafka.go
```diff
@@ -32,6 +32,18 @@ var maxWaitTimeToSendMSG = time.Second * 30
 
 var _ Syncer = &KafkaSyncer{}
 
+// PartitionMode Kafka partition mode
+type PartitionMode byte
+
+const (
+	// PartitionFixed choose fixed partition 0
+	PartitionFixed = iota + 1
```
kennytm:
```suggestion
	PartitionFixed PartitionMode = iota + 1
```

drainer/sync/kafka.go
```diff
@@ -112,12 +142,60 @@ func (p *KafkaSyncer) Sync(item *Item) error {
...
+		}
+	}
+	return nil
+}
+
+func (p *KafkaSyncer) syncDMLPartitionByTable(binlog *obinlog.Binlog, item *Item) error {
+	tables := binlog.DmlData.Tables
+	if len(tables) == 1 {
+		table := tables[0]
+		return errors.Trace(p.saveBinlog(binlog, item, sarama.StringEncoder(*table.SchemaName+"."+*table.TableName)))
```
kennytm: This specialization doesn't seem useful. The outcome of `len(tables) == 1` and `len(tables) != 1` are exactly the same.

sunxiaoguang: This can save one allocation. But it's small though and probably can be ignored.

drainer/sync/kafka.go
sunxiaoguang: Fixed

PR #585, opened 2019-05-07
pkg/loader: Refactor and add tests for translate functions

suzaku:

[Related Issue](https://internal.pingcap.net/jira/browse/TOOL-1050)

Improve test coverage: 39.1% -> 45.3%

### Check List 

Tests 

 - Unit test

Code changes

Side effects

Related changes

### DIFF
--- a/pkg/loader/translate_test.go
+++ b/pkg/loader/translate_test.go
@@ -97,6 +97,18 @@ func (s *slaveBinlogToTxnSuite) TestTranslateDML(c *C) {
 	c.Assert(insert.Values["uid"], Equals, newVal)
 }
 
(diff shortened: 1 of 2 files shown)


pkg/loader/translate_test.go
```diff
@@ -0,0 +1,134 @@
...
+							},
+						},
+						ChangeRow: &pb.Row{
+							Columns: []*pb.Column{
+								{Int64Value: &oldVal},
+							},
+						},
+					},
+					{
+						Type: pb.MutationType_Insert.Enum(),
```
csuzhangxc: add case for `DeleteDMLType` and `UnknownDMLType` too?

suzaku: If we want, we can add those two in tests for `getDMLType`.

PR #593, opened 2019-05-09
drainer: Ignore empty string when getting compressor config

suzaku:

Although empty compressor name seems to be ignored by grpc anyway, we'd
better not depend on undocumented behaviors.

Make sure empty compressors are ignored.

### Check List 

Tests 

 - Unit test

Code changes

Side effects

Related changes

### DIFF
--- a/drainer/pump.go
+++ b/drainer/pump.go
@@ -14,6 +14,7 @@
 package drainer
 
 import (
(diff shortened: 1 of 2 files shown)


WangXiangUSTC:
LGTM

PR #595, opened 2019-05-09
Turn fatal error listening on 127.0.0.1/localhost into a warning (was #561)

kolbe:
Drainer logs a fatal error when it encounters an attempt to bind to 127.0.0.1 or localhost. The concern seems to be that pump will not be able to contact the drainer in this case? This is not a very reasonable assumption, since someone might run a local cluster for testing or use port forwarding.

This PR changes the fatal error to a warning.

### DIFF
--- a/drainer/config.go
+++ b/drainer/config.go
@@ -234,7 +234,7 @@ func (cfg *Config) validate() error {
 	}
 
 	if !util.IsValidateListenHost(host) {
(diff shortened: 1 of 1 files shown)


kolbe:
I believe all other TiDB Platform components allow binding to loopback addresses. I don't see why drainer should be an exception!

amyangfei:
@july2993 I have no write permission to this repo

PR #596, opened 2019-05-10
*: Create a reusable metricClient in util

suzaku:

Eliminate duplicate code for starting a goroutine that push metrics.

Extract the common code to util and remove the original duplicate
implementation.

### Check List 

Tests 

 - Unit test

Code changes

Side effects

Related changes

### DIFF
--- a/pkg/util/p8s.go
+++ b/pkg/util/p8s.go
@@ -1,0 +1,58 @@
+// Copyright 2019 PingCAP, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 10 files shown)


pkg/util/p8s.go
```diff
@@ -0,0 +1,58 @@
...
+
+// MetricClient manage the periodic push to the Prometheus Pushgateway.
+type MetricClient struct {
+	addr     string
+	interval time.Duration
+	registry *prometheus.Registry
+}
+
+// Start run a loop of pushing metrics to Prometheus Pushgateway.
+func (mc MetricClient) Start(ctx context.Context, grouping map[string]string) {
```
WangXiangUSTC: seems all the grouping is `"instance": nodeid`, how about use nodeid as arg

suzaku: @july2993 What's your opinion? Do we want more flexibility here or shorter code?

july2993: i prefer keep `grouping`

WangXiangUSTC:
LGTM

PR #600, opened 2019-05-13
*: Fix some missing error checks as reported by golangci-lint

suzaku:

Fix some missing error checks as reported by golangci-lint

### Check List 

Tests 

 - Unit test

Code changes

Side effects

Related changes

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,3 +33,4 @@ cscope.*
 
 # Files generated when testing
 out
+vendor/
--- a/arbiter/config.go
+++ b/arbiter/config.go
@@ -137,7 +137,9 @@ func (cfg *Config) Parse(args []string) error {
 	}
 
 	// parse again to replace with command line options
-	cfg.FlagSet.Parse(args)
+	if err := cfg.FlagSet.Parse(args); err != nil {
+		return errors.Trace(err)
+	}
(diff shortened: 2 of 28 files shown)


july2993:
pls fix ci, and this ps fix `some` not `all`?

suzaku:
> pls fix ci, and this ps fix `some` not `all`?

Yes, this PR only want to fix some of the easy ones.

The others are more complex, some involves changing the workflow and others are just result of bad design of the library API.
For the first kind, we can fix them in separate PRs instead of hiding them among other simple fixes.
For the second kind, I don't think we want to fix them at all.

WangXiangUSTC:
LGTM

pump/storage/log.go
```diff
@@ -338,7 +338,13 @@ func seekToNextRecord(reader *bufio.Reader) (bytes int, err error) {
 			return
 		}
 
-		reader.Discard(1)
+		var discarded int
+		discarded, err = reader.Discard(1)
+		if err != nil {
```
july2993: At L329
we `reader.Peek(4)` and success, so we should always success to Discard one here.
what about just fatal if `err != nil` here and add some comment

suzaku: I'm not sure. What's the difference it should never happens?

PR #608, opened 2019-05-20
go.mod: Update parser dep

july2993:

update to the version which tidb 2.1 use.
before this pr, fail to parser this SQL:
create table a(id int, a varchar(30) NOT NULL COMMENT 'aa' COLLATE
	'utf8_bin');
`GO111MODULE=on go get github.com/pingcap/parser@361d2d4f774d`

### Check List 

Tests 

run the sql as desc before

Code changes

Side effects

Related changes


suzaku:
LGTM

PR #611, opened 2019-05-21
load.go: set max idle conns as worker count

july2993:
the default value is only 2, and may change for future release of go
when the worker count if hight, this will cause two many connection
create and close.

set max idle conns as worker count

the default value is only 2, and may change for future release of go
when the worker count if hight, this will cause two many connection
create and close every time we use `sql.DB`.

set max idle conns as worker count

### Check List 

Tests 
worker-count = 64

before this pr, it may create and close connection frequencly, and not port available more.
need wait `TIME_WAIT` before reuse port.
```
[2019/05/21 10:17:49.618 +08:00] [ERROR] [util.go:147] [singleExec] [error="dial tcp 172.16.4.71:3306: connect: cannot assign requested address"] [errorVerbose="dial tcp 172.16.4.71:3306: connect: cannot assign requested address\ngithub.com/pingcap/errors.AddStack\n\t/home/jenkins/workspace/release_tidb_3.0/go/pkg/mod/github.com/pingcap/errors@v0.11.0/errors.go:174\ngithub.com/pingcap/errors.Trace\n\t/home/jenkins/workspace/release_tidb_3.0/go/pkg/mod/github.com/pingcap/errors@v0.11.0/juju_adaptor.go:12\ngithub.com/pingcap/tidb-binlog/pkg/loader.(*executor).begin\n\t/home/jenkins/workspace/release_tidb_3.0/go/src/github.com/pingcap/tidb-binlog/pkg/loader/executor.go:107\ngithub.com/pingcap/tidb-binlog/pkg/loader.(*executor).singleExec\n\t/home/jenkins/workspace/release_tidb_3.0/go/src/github.com/pingcap/tidb-binlog/pkg/loader/executor.go:261\ngithub.com/pingcap/tidb-binlog/pkg/loader.(*executor).singleExecRetry.func1\n\t/home/jenkins/workspace/release_tidb_3.0/go/src/github.com/pingcap/tidb-binlog/pkg/loader/executor.go:250\ngithub.com/pingcap/tidb-binlog/pkg/util.RetryOnError\n\t/home/jenkins/workspace/release_tidb_3.0/go/src/github.com/pingcap/tidb-binlog/pkg/util/util.go:142\ngithub.com/pingcap/tidb-binlog/pkg/loader.(*executor).singleExecRetry\n\t/home/jenkins/workspace/release_tidb_3.0/go/src/github.com/pingcap/tidb-binlog/pkg/loader/executor.go:249\ngithub.com/pingcap/tidb-binlog/pkg/loader.(*loaderImpl).execByHash.func1\n\t/home/jenkins/workspace/release_tidb_3.0/go/src/github.com/pingcap/tidb-binlog/pkg/loader/load.go:292\ngolang.org/x/sync/errgroup.(*Group).Go.func1\n\t/home/jenkins/workspace/release_tidb_3.0/go/pkg/mod/golang.org/x/sync@v0.0.0-20180314180146-1d60e4601c6f/errgroup/errgroup.go:58\nruntime.goexit\n\t/usr/local/go/src/runtime/asm_amd64.s:1337"]
```

after this pr: for insert stably 70+ K/s
Screen Shot 2019-05-21 at 1 29 33 PM

Code changes

Side effects

Related changes

### DIFF
--- a/pkg/loader/load.go
+++ b/pkg/loader/load.go
@@ -142,6 +142,7 @@ func NewLoader(db *gosql.DB, opt ...Option) (Loader, error) {
 	}
 
 	db.SetMaxOpenConns(opts.workerCount)
+	db.SetMaxIdleConns(opts.workerCount)
 
 	return s, nil
 }


pkg/loader/load.go
```diff
@@ -142,6 +142,7 @@ func NewLoader(db *gosql.DB, opt ...Option) (Loader, error) {
 	}
 
 	db.SetMaxOpenConns(opts.workerCount)
+	db.SetMaxIdleConns(opts.workerCount)
```
suzaku: How about setting it to a lower value so that we don't need to keep many idle connections when drainer's not busy?

july2993: should not do this, or you will need to create new connection every time the current use of `sql.DB` is bigger than the value you set.
but https://godoc.org/database/sql#DB.SetConnMaxLifetime can get what you want.

suzaku: I see. Your assumption is that `workerCount` will be fully used most of the time.
What I'm thinking about is if that is not the case, we would be wasting memory of drainer and the downstream database just to store connections that are going to timeout anyway.

july2993: how's the cost of memory of for one conn? and how's the cost you create new connection every time?

suzaku: I mean it depends on the use ratio of `workerCount`.

GregoryIan:
the monitor can not say more, can you past the specific value of promotion? @july2993

GregoryIan:
LGTM

july2993:
> 1. if `MaxOpenConns` can be very large (such as 20k, 40k, or a mass of applications connecting to the same db are deployed), I suggest to set two variables, one is `MaxOpenConns`, the other is `MaxIdleConns`, for the following considerations:
>    
>    * too many idle connections cost more memory
>    * too many db connections may cause too many reconnect behaviors after db max_wait_timeout
> 2. If `MaxOpenConns` is not very large, keep as many idle connections as max allowed open connections is good for scenario in this issue.

`MaxOpenConns` will not be very large and will be the same value of `worker-count`, like in `dm syncer`,  but it create worker-count * `sql.DB`, will not get any benefit for such as 20k, 40k, i think it's ok the keep it as max allowed open connections now.

PR #612, opened 2019-05-21
tests: Stop pumps one by one to avoid stopping the upstream TiDB

suzaku:

In the `restart` test, we tried to restart 2 pump instances at almost the same time with background jobs.
Meanwhile, another background job keeps inserting data into a test
table.
When prewriting a binlog for the inserted data and no pump instances is
available, TiDB would fail with critical error.

Stop pumps synchronously.

### Check List 

Tests 

 - Integration test

Code changes

Side effects

Related changes

### DIFF
--- a/tests/_utils/run_pump
+++ b/tests/_utils/run_pump
@@ -9,7 +9,7 @@ OUT_DIR=/tmp/tidb_binlog_test
 # kill pump, util no pump process is running
 while :
 do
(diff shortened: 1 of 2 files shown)


WangXiangUSTC:
@july2993 yes，I think we should not modify the test, if the test failed, we should check the pump client

GregoryIan:
@suzaku is it reviewable?

suzaku:
Need to reproduce the bug and fix it instead of changing the test.

PR #613, opened 2019-05-21
loader/: don't refresh table info for some DDL

july2993:
currently for after every ddl we execute, we call refreshTableInfo, but for some ddl, no need the do this.
like for `drop table test` in database test, we call `refreshTableInfo("test", "test")`, but the downstream table has been dropped.

some relate log
```
[2019/05/21 16:44:37.390 +08:00] [INFO] [load.go:270] ["exec ddl success"] [sql="DROP TABLE IF EXISTS sbtest18"]
[2019/05/21 16:44:37.390 +08:00] [INFO] [syncer.go:241] ["write save point"] [ts=408434837891317763]
[2019/05/21 16:44:37.391 +08:00] [WARN] [load.go:215] ["table has no any primary key and unique index, it may be slow when syncing data to downstream, we highly recommend add primary key or unique key for table"] [table=`sbtest`.`sbtest18`]
```

don't call refreshTableInfo for some ddl
also make `getTableInfo` return err if the table not exist instead of return a invalid table info

### Check List 

Tests 

 - Unit test
run and check log
 Code changes

Side effects

Related changes


pkg/loader/load_test.go
```diff
@@ -236,6 +236,23 @@ func (s *isCreateDBDDLSuite) TestCreateDBSQL(c *check.C) {
 	c.Assert(isCreateDatabaseDDL("create database `db2`;"), check.IsTrue)
 }
 
+type needRefreshTableInfoSuite struct{}
+
+var _ = check.Suite(&needRefreshTableInfoSuite{})
+
+func (s *needRefreshTableInfoSuite) TestNeedRefreshTableInfo(c *check.C) {
+	cases := map[string]bool{
+		"DROP TABLE a":           false,
```
suzaku: Maybe we can use `IsFalse` and simplify the `Assert` call below?

pkg/loader/load.go
```diff
@@ -229,6 +229,25 @@ func (s *loaderImpl) getTableInfo(schema string, table string) (info *tableInfo,
...
+		return false
+	}
+
+	switch stmt.(type) {
+	case *ast.DropTableStmt:
+		return false
+	case *ast.DropDatabaseStmt:
+		return false
+	case *ast.TruncateTableStmt:
+		return false
```
suzaku: How about merging them in one case? It seems like we are doing that in other places like `IgnoreDDLError`.

WangXiangUSTC:
LGTM

GregoryIan:
@suzaku PTAL

pkg/loader/load.go
```diff
@@ -211,6 +214,10 @@ func (s *loaderImpl) refreshTableInfo(schema string, table string) (info *tableI
 		return info, errors.Trace(err)
 	}
 
+	if len(info.columns) == 0 {
```
suzaku: It would be more clear here if we have a comment to explain why empty `info.columns` means and only means "table not exist".

pkg/loader/util.go
```diff
@@ -171,6 +179,12 @@ func getColsOfTbl(db *gosql.DB, schema, table string) ([]string, error) {
 	if err = rows.Err(); err != nil {
 		return nil, errors.Trace(err)
 	}
+
+	// if no any columns returns, means the table not exist.
```
suzaku: IMHO, directly testing whether there's no `rows` returned for the specified schema and table name would be clearer.

july2993: I think we should use https://godoc.org/database/sql#DB.QueryRow

july2993: what do you means no rows return?
when there empty rows result.
will also return rows, and err = nil.
but the Next() will return false at first time you call it `rows.Next()`

suzaku: Is there any other cases when the table exists but `cols` is empty? If no, LGTM.

july2993:

pkg/loader/util.go
july2993: no other cases

pkg/loader/load.go
kennytm: Why is this marked "resolved"? There are still 4 cases which can be merged into one.

july2993:

PR #614, opened 2019-05-21
tests: Support selecting multiple cases to run

suzaku:

We can't select more than one integration test case to run.

Support selecting multiple cases to run.

### Check List 

Tests 

 - Integration test

Code changes

Side effects

Related changes

### DIFF
--- a/tests/run.sh
+++ b/tests/run.sh
@@ -127,21 +127,30 @@ if [ "${1-}" = '--debug' ]; then
     read line
 fi
 
-# List the case names to run, separate them by space, eg. "binlog gencol"
-do_cases=""
-
-for script in ./*/run.sh; do
-    test_name="$(basename "$(dirname "$script")")"
-    if [[ $do_cases != "" && ! $do_cases =~ (^|[[:space:]])$test_name($|[[:space:]]) ]]; then
-        continue
-    fi
-
+run_case() {
+    local case=$1
+    local script=$2
     echo "Running test $script..."
     PATH="$pwd/../bin:$pwd/_utils:$PATH" \
     OUT_DIR=$OUT_DIR \
-    TEST_NAME=$test_name \
+    TEST_NAME=$case \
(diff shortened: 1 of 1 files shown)


tests/run.sh
```diff
@@ -127,12 +127,12 @@ if [ "${1-}" = '--debug' ]; then
...
 
-# set to the case name you want to run only for debug
-do_case=""
+# List the case names to run, separate them by space, eg. "binlog gencol"
+do_cases=""
 
 for script in ./*/run.sh; do
     test_name="$(basename "$(dirname "$script")")"
-    if [[ $do_case != "" && $test_name != $do_case ]]; then
+    if [[ $do_cases != "" && ! $do_cases =~ (^|[[:space:]])$test_name($|[[:space:]]) ]]; then
```
amyangfei: note `[[` is not available in `sh`, and bash is superset of sh, it is better to change the shebang to `#!/bin/bash`

tests/run.sh
```diff
@@ -127,12 +127,12 @@ if [ "${1-}" = '--debug' ]; then
     read line
 fi
 
-# set to the case name you want to run only for debug
-do_case=""
+# List the case names to run, separate them by space, eg. "binlog gencol"
+do_cases=""
```
WangXiangUSTC: how about use array in shell  https://stackoverflow.com/questions/1878882/arrays-in-unix-shell

suzaku: Is there an easy way to test if a case is among the selected ones with array?

WangXiangUSTC: if do_cases' length is not 0, just run code like:
```
for case in $do_cases; do
      sh tests/$case/run.sh
done
```

WangXiangUSTC:
LGTM

suzaku:
@amyangfei PTAL

GregoryIan:
@amyangfei PTAL

PR #615, opened 2019-05-22
*: Simplify checkpoint

july2993:
Simplify checkpoint

Drop the Check() and and String() of Checkpoint.
Drop MetaCheckpoint of FlashCheckPoint.

After https://github.com/pingcap/tidb-binlog/pull/532
we make introduce a interface
```
// Syncer sync binlog item to downstream
type Syncer interface {
	// Sync the binlog item to downstream
	Sync(item *Item) error
	// will be close if Close normally or meet error, call Error() to check it
	Successes() <-chan *Item
	// Return not nil if fail to sync data to downstream or nil if closed normally
	Error() <-chan error
	// Close the Syncer, no more item can be added by `Sync`
	Close() error
}
```
if make sure if the binlog item return in `Successes()`, then this item and all previous item has been 
write to downstream.

the old code using a `executor` to executor the sql(not really sql for flash downstream), but internally the flash executor will do some batch(means not write the data to downstream and just return success when call `execute(..)`, and flush data to downstream async.
it use the global `metaCP` in `pkg/flash` to make sure `checkpoint` not save a not really safe timestamp.

There's no need for the anymore.

### Check List 

Tests 

 - Unit test
 - Integration test

Side effects

 Related changes


suzaku:
LGTM

PR #617, opened 2019-05-23
pump: getSafeGCTSOForDrainers should be usable after Close

suzaku:

When we call `Close` on the server, it first calls `cancel`
to stop all related background jobs. If we reuse `s.ctx` in
`getSafeGCTSOForDrainers` after `Close`, nothing useful can be done.

Make `ctx` a parameter, so that when calling `getSafeGCTSOForDrainers`
after `Close` we can pass in a different context.

### Check List 

Tests 

 - Integration test

Code changes

Side effects

Related changes

### DIFF
--- a/pump/server.go
+++ b/pump/server.go
@@ -500,7 +500,7 @@ func (s *Server) gcBinlogFile() {
 				continue
 			}
 
-			safeTSO, err := s.getSafeGCTSOForDrainers()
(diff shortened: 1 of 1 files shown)


july2993:
LGTM

suzaku:
集成测试失败原因，跑 kafka case 时 TiKV server timeout

WangXiangUSTC:
LGTM

PR #621, opened 2019-05-27
reparo: refine integration test

WangXiangUSTC:
reparo's integration test is too simple, should refine it

1. use daily test in reparo's integration test
2. find a bug, after exec DDL, loader will refresh table info, but the DDL.Database and DDL.Table is empty, and the DML will execute failed because use old table info.

### Check List 

Tests 
 - Integration test

Related changes

### DIFF
--- a/pkg/loader/load.go
+++ b/pkg/loader/load.go
@@ -493,7 +493,7 @@ func newBatchManager(s *loaderImpl) *batchManager {
 		fDDLSuccessCallback: func(txn *Txn) {
 			s.markSuccess(txn)
 			if _, err := s.refreshTableInfo(txn.DDL.Database, txn.DDL.Table); err != nil {
-				log.Errorf("refresh table info failed, %v", err)
+				log.Error("refresh table info failed", zap.Error(err))
 			}
 		},
 	}
--- a/pkg/loader/load.go
+++ b/pkg/loader/load.go
@@ -545,6 +545,10 @@ func (b *batchManager) put(txn *Txn) error {
 	// we always executor the previous dmls when we meet ddl,
 	// and executor ddl one by one.
 	if txn.isDDL() {
+		if len(txn.DDL.Database) == 0 {
+			return errors.Errorf("get DDL Txn with empty database, ddl: %s", txn.DDL.SQL)
+		}
+
 		if err := b.execAccumulatedDMLs(); err != nil {
 			return errors.Trace(err)
 		}
--- a/pkg/loader/load.go
+++ b/pkg/loader/load.go
@@ -206,6 +206,8 @@ func (s *loaderImpl) Close() {
 var utilGetTableInfo = getTableInfo
 
 func (s *loaderImpl) refreshTableInfo(schema string, table string) (info *tableInfo, err error) {
+	log.Info("refresh table info", zap.String("schema", schema), zap.String("table", table))
+
 	if len(schema) == 0 {
 		return nil, errors.New("schema is empty")
 	}
--- a/pkg/loader/load.go
+++ b/pkg/loader/load.go
@@ -495,7 +495,7 @@ func newBatchManager(s *loaderImpl) *batchManager {
 		fDDLSuccessCallback: func(txn *Txn) {
 			s.markSuccess(txn)
 			if _, err := s.refreshTableInfo(txn.DDL.Database, txn.DDL.Table); err != nil {
-				log.Error("refresh table info failed", zap.Error(err))
+				log.Error("refresh table info failed", zap.String("database", txn.DDL.Database), zap.String("table", txn.DDL.Table), zap.Error(err))
 			}
 		},
 	}
--- a/pkg/loader/load.go
+++ b/pkg/loader/load.go
@@ -206,6 +206,14 @@ func (s *loaderImpl) Close() {
 var utilGetTableInfo = getTableInfo
 
 func (s *loaderImpl) refreshTableInfo(schema string, table string) (info *tableInfo, err error) {
+	if len(schema) == 0 {
+		return nil, errors.New("schema is empty")
+	}
+
+	if len(table) == 0 {
+		return nil, nil
+	}
+
 	info, err = utilGetTableInfo(s.db, schema, table)
 	if err != nil {
 		return info, errors.Trace(err)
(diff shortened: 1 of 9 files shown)


pkg/loader/load.go
```diff
@@ -206,6 +206,14 @@ func (s *loaderImpl) Close() {
 var utilGetTableInfo = getTableInfo
 
 func (s *loaderImpl) refreshTableInfo(schema string, table string) (info *tableInfo, err error) {
+	if len(schema) == 0 {
```
GregoryIan: why schema would be empty? what's the case?

pkg/loader/load.go
```diff
@@ -484,7 +492,9 @@ func newBatchManager(s *loaderImpl) *batchManager {
 		fExecDDL:             s.execDDL,
 		fDDLSuccessCallback: func(txn *Txn) {
 			s.markSuccess(txn)
```
GregoryIan: mark success but fail to refresh, does it have any effect？

GregoryIan:
What is its relationship with #613 ?

reparo/syncer/translate.go
```diff
@@ -29,6 +32,10 @@ func pbBinlogToTxn(binlog *pb.Binlog) (txn *loader.Txn, err error) {
 		txn.DDL = new(loader.DDL)
 		// for table DDL, pb.Binlog.DdlQuery will be "use <db>; create..."
 		txn.DDL.SQL = string(binlog.DdlQuery)
+		txn.DDL.Database, txn.DDL.Table, err = parserSchemaTableFromDDL(txn.DDL.SQL)
```
GregoryIan: I think we should check `txn.DDL.Database` and `txn.DDL.Table` here

GregoryIan:
it's really a simple and sick bug 🤢

reparo/syncer/translate.go
```diff
@@ -137,3 +144,63 @@ func genColsAndArgs(row [][]byte) (cols []string, args []interface{}, err error)
 
 	return
 }
+
+// parserSchemaTableFromDDL parse ddl query to get schema and table
```
suzaku:
```suggestion
// parserSchemaTableFromDDL parses ddl query to get schema and table
```

reparo/syncer/translate.go
```diff
@@ -137,3 +144,63 @@ func genColsAndArgs(row [][]byte) (cols []string, args []interface{}, err error)
...
+// ddl like use test; create table ....
```
suzaku:
```suggestion
// ddl like `use test; create table`
```

reparo/syncer/translate.go
```diff
@@ -137,3 +144,63 @@ func genColsAndArgs(row [][]byte) (cols []string, args []interface{}, err error)
...
+func parserSchemaTableFromDDL(ddlQuery string) (schema, table string, err error) {
+	stmts, _, err := parser.New().Parse(ddlQuery, "", "")
+	if err != nil {
+		return "", "", err
+	}
+
+	for _, stmt := range stmts {
```
suzaku: Is it possible that `stmts` contains more than 2 statements and involves more than 1 table?

WangXiangUSTC: should not have this situation

pkg/loader/load.go
WangXiangUSTC: user may forget set the DDL.Database

reparo/syncer/translate.go
WangXiangUSTC: done

pkg/loader/load.go
WangXiangUSTC: mark success used for save checkpoint, if refreshTableInfo failed, only influence the dmls after this ddl.  I think it's ok to markSuccess here.

suzaku: It sent the transaction to the channel returned by `Successes()`, which is used in `Syncer.handleSuccess` and is mainly used to parse for timestamp and be saved as checkpoints.
Since at this point the DDL is already synced to the downstream database and `refreshTableInfo` only affect in memory status of drainer, it should be OK.

reparo/syncer/translate.go
```diff
@@ -137,3 +147,63 @@ func genColsAndArgs(row [][]byte) (cols []string, args []interface{}, err error)
...
+			if len(node.Table.Schema.O) != 0 {
+				schema = node.Table.Schema.O
+			}
+			table = node.Table.Name.O
+		case *ast.AlterTableStmt:
+			if len(node.Table.Schema.O) != 0 {
+				schema = node.Table.Schema.O
+			}
+			table = node.Table.Name.O
+		case *ast.DropTableStmt:
```
suzaku: It seems that the parsings of operations that only involve 1 table are identical, maybe we can merge them into 1 case to simplify this function?

WangXiangUSTC: can't merge them, because the stmt interface don't have `Table`, or can you provide some example code?

suzaku: Like this?

```go
		case *ast.DropIndexStmt, *ast.AlterTableStmt:
			if len(node.Table.Schema.O) != 0 {
				schema = node.Table.Schema.O
			}
			table = node.Table.Name.O
```

WangXiangUSTC: can't do this, the interface don't have Table, will get error `type ast.StmtNode has no field or method Table`

reparo/syncer/translate.go
GregoryIan: should we check this case and return error?

reparo/syncer/translate.go
suzaku: I see, typical Go ...

reparo/syncer/translate.go
suzaku: IMHO it would be safer if we check to make sure it doesn't involve more than 1 table.

pkg/loader/load.go
```diff
@@ -484,7 +494,9 @@ func newBatchManager(s *loaderImpl) *batchManager {
...
-			s.refreshTableInfo(txn.DDL.Database, txn.DDL.Table)
+			if _, err := s.refreshTableInfo(txn.DDL.Database, txn.DDL.Table); err != nil {
+				log.Error("refresh table info failed", zap.Error(err))
```
july2993: add info about `txn.DDL.Database` `txn.DDL.Table`?

WangXiangUSTC:
@suzaku forget set DDL.Database and DDL.Table, after exectue ddl, loader need refresh table info by query mysql/tidb, but if  DDL.Database and DDL.Table is empty, can not get the new table info. so the dmls will use the old schema, and execute failed

reparo/syncer/translate.go
WangXiangUSTC: only `drop table` and `rename table` may involve more than one table, but this seems is not a serious problem now, will fix it later

WangXiangUSTC:
will add other test later @july2993

pkg/loader/load.go
WangXiangUSTC: done

PR #622, opened 2019-05-28
drainer: Log close error and long waits

suzaku:

[Related Issue](https://internal.pingcap.net/jira/browse/TOOL-1235?filter=-2)
The failure of `s.dsyncer.Close` may cause the following
`wg.Wait()` to block forever, we don't even have a chance to look at the
returned error.

Log the close error before starting to block waiting, and log long
waits.

### Check List 

Tests 

 - Unit test

Code changes

Side effects

Related changes

### DIFF
--- a/drainer/syncer.go
+++ b/drainer/syncer.go
@@ -406,8 +406,7 @@ ForLoop:
 	select {
 	case <-wait:
 	case <-time.After(runWaitThreshold):
-		log.Error("Waiting too long for `Syncer.run` to quit.")
-		<-wait
+		panic("Waiting too long for `Syncer.run` to quit.")
 	}
 
 	close(s.closed)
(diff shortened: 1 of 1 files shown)


WangXiangUSTC:
LGTM

drainer/syncer.go
```diff
@@ -396,7 +399,16 @@ ForLoop:
...
-	wg.Wait()
+	if cerr != nil {
+		log.Error("Failed to close syncer", zap.Error(cerr))
+	}
+
+	select {
+	case <-wait:
+	case <-time.After(runWaitThreshold):
+		log.Error("Waiting too long for `Syncer.run` to quit.")
+		<-wait
```
GregoryIan: we can remove it, force it exit

suzaku: OK, do you think 10 seconds is long enough?

GregoryIan: enough in general

GregoryIan:
LGTM

PR #624, opened 2019-05-29
pump/storage: Random cleanups

suzaku:

Clean up come code in pump/storage.

1. Use predefined constants in leveldb/opt to improve readability of
   leveldb config values.
1. Use path/filepath and strings functions to simplify path manipulation
1. Rename confusing variable

### Check List 

Tests 

 - Unit test

Code changes

Side effects

Related changes

### DIFF
--- a/pump/storage/storage.go
+++ b/pump/storage/storage.go
@@ -1054,14 +1054,14 @@ type KVConfig struct {
 }
 
 var defaultStorageKVConfig = &KVConfig{
-	BlockCacheCapacity:            8388608,
+	BlockCacheCapacity:            8 * opt.MiB,
 	BlockRestartInterval:          16,
-	BlockSize:                     4096,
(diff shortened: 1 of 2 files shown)


pump/storage/storage.go
```diff
@@ -732,26 +732,26 @@ func (a *Append) writeToValueLog(reqs chan *request) chan *request {
 		var bufReqs []*request
 		var size int
 
-		write := func(reqs []*request) {
-			if len(reqs) == 0 {
+		write := func(batch []*request) {
```
WangXiangUSTC: the value's type is request, maybe reqs is better, I think don't need change value name

suzaku: The name `reqs` is already taken above for the channel.

GregoryIan:
LGTM

PR #625, opened 2019-05-30
tests: fix the value range when checking exec txn histogram

amyangfei:
In `ExecuteTxn` test, we execute two SQLs with some delay (101ms, 201ms) via sql mock.
the original check range for histogram sample_sum is [0.302, 0.357], these two metrics bound reflect the delay from `txn.Exec(...)` to `histogram.WithLabelValues("exec").Observe(...)`, from  302ms to 357ms.
We can have a more accurate upper bound based on execution time observing from test case. (sometime the execution time may even exceed 357ms)

change the check range to (0.302, observing upper bound)

### Check List 

Tests 

 - Unit test

### DIFF
--- a/pkg/sql/sql_test.go
+++ b/pkg/sql/sql_test.go
@@ -126,20 +126,29 @@ const (
 )
 
 func (s *sqlSuite) TestExecuteTxnSuccess(c *C) {
+	var (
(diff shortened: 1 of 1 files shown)


GregoryIan:
LGTM! Good job

PR #629, opened 2019-06-03
*: empty host should not be considered valid, fix #627

### LINKED ISSUE #627: Validation of `-addr` and `-advertise-addr`
When given `-addr :8250` without specifying `-advertise-addr`, both of these options end up being ":8250".

Validation should be added so that invalid address that can't be connected should fail the check.

suzaku:

Fix #627

1. Reuse `IsValidateListenHost` for checking `advertiseHost`
1. Make sure empty string is invalid

### Check List 

Tests 

 - Unit test
 - Manual test

Code changes

Side effects

Related changes


WangXiangUSTC:
LGTM

PR #631, opened 2019-06-06
pump: Add config to set cap buf of kv_chan

july2993:
let user to set 
* the kv channel capacity to buffer  more binlog meta
* output binlog meta information if pump take a long time to write binlog

add configuration items for pump storage under [storage] configuration

``` 
# the channel to buffer binlog meta, pump will block write binlog request if the channel is full
# kv_chan_cap = 4194304
#
# if pump takes a long time to write binlog, pump will display the binlog meta information
# slow_write_threshold = 2
```

### Check List 

Tests 

### DIFF
--- a/cmd/pump/pump.toml
+++ b/cmd/pump/pump.toml
@@ -31,6 +31,7 @@ pd-urls = "http://127.0.0.1:2379"
 # Set to `true` (default) for best reliability, which prevents data loss when there is a power failure.
 # sync-log = true
 #
+# kv_chan_cap = 0
 # we suggest using the default config of the embedded LSM DB now, do not change it useless you know what you are doing
 # [storage.kv]
 # block-cache-capacity = 8388608
--- a/cmd/pump/pump.toml
+++ b/cmd/pump/pump.toml
@@ -35,6 +35,7 @@ pd-urls = "http://127.0.0.1:2379"
 # kv_chan_cap = 4194304
 #
 # if pump takes a long time to write binlog, pump will display the binlog meta information
+# now the unit is `Second`
 # slow_write_duration = 2
 #
 # we suggest using the default config of the embedded LSM DB now, do not change it useless you know what you are doing
--- a/cmd/pump/pump.toml
+++ b/cmd/pump/pump.toml
@@ -31,13 +31,6 @@ pd-urls = "http://127.0.0.1:2379"
 # Set to `true` (default) for best reliability, which prevents data loss when there is a power failure.
 # sync-log = true
 #
-# the channel to buffer binlog meta, pump will block write binlog request if the channel is full
-# kv_chan_cap = 4194304
-#
-# if pump takes a long time to write binlog, pump will display the binlog meta information
-# now the unit is `Second`
-# slow_write_duration = 2
-#
 # we suggest using the default config of the embedded LSM DB now, do not change it useless you know what you are doing
 # [storage.kv]
 # block-cache-capacity = 8388608
--- a/cmd/pump/pump.toml
+++ b/cmd/pump/pump.toml
@@ -31,7 +31,12 @@ pd-urls = "http://127.0.0.1:2379"
 # Set to `true` (default) for best reliability, which prevents data loss when there is a power failure.
 # sync-log = true
 #
-# kv_chan_cap = 0
+# the channel to buffer binlog meta, pump will block write binlog request if the channel is full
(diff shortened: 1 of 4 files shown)


GregoryIan:
LGTM

cmd/pump/pump.toml
```diff
@@ -31,6 +31,7 @@ pd-urls = "http://127.0.0.1:2379"
 # Set to `true` (default) for best reliability, which prevents data loss when there is a power failure.
 # sync-log = true
 #
+# kv_chan_cap = 0
```
WangXiangUSTC: how about set a default value

WangXiangUSTC:
rest LGTM

cmd/pump/pump.toml
WangXiangUSTC: and need add a comment

pump/storage/storage.go
```diff
@@ -611,10 +612,15 @@ func (a *Append) writeBinlog(binlog *pb.Binlog) *request {
...
 	defer func() {
-		writeBinlogTimeHistogram.WithLabelValues("single").Observe(time.Since(beginTime).Seconds())
+		duration := time.Since(beginTime).Seconds()
+		writeBinlogTimeHistogram.WithLabelValues("single").Observe(duration)
 		if request.err != nil {
 			errorCount.WithLabelValues("write_binlog").Add(1.0)
 		}
+
+		if duration > a.options.SlowWriteDuration {
+			log.Warn("take a long time to write binlog", zap.Stringer("binlog type", binlog.Tp), zap.Int64("commit TS", binlog.StartTs), zap.Int64("start TS", binlog.CommitTs), zap.Int("length", len(binlog.PrewriteValue)))
```
july2993: switch  name `zap.Int64("commit TS", binlog.StartTs), zap.Int64("start TS", binlog.CommitTs)`

cmd/pump/pump.toml
```diff
@@ -31,6 +31,12 @@ pd-urls = "http://127.0.0.1:2379"
 # Set to `true` (default) for best reliability, which prevents data loss when there is a power failure.
 # sync-log = true
 #
+# the channel to buffer binlog meta, pump will block write binlog request if the channel is full
+# kv_chan_cap = 4194304
+#
+# if pump takes a long time to write binlog, pump will display the binlog meta information
+# slow_write_duration = 2
```
july2993: better use a name **second, if you expect it as second

GregoryIan: I add a comment and don't change name, because I do't want to change the name if we change the definition of precision

cmd/pump/pump.toml
GregoryIan: added

pump/storage/storage.go
```diff
@@ -40,7 +40,10 @@ import (
 
 const (
 	maxTxnTimeoutSecond int64 = 600
-	chanSize                  = 1 << 20
+	// the channel to buffer binlog meta, pump will block write binlog request if the channel is full
+	chanSize = 1 << 20
+	// if pump takes a long time to write binlog, pump will display the binlog meta information (unit: Second)
+	slowWriteDuration = 1.0
```
suzaku: Threshold would be clearer, for example, a similar concept is MySQL's query time threshold.

pump/storage/storage.go
```diff
@@ -135,7 +138,7 @@ func NewAppendWithResolver(dir string, options *Options, tiStore kv.Storage, tiL
 		return nil, errors.Trace(err)
 	}
 
-	writeCh := make(chan *request, chanSize)
+	writeCh := make(chan *request, options.KVChanCap)
```
suzaku: It seems like this is the input channel for `writeValueLog`, but the name `KVChanCap` suggests that it's for the output channel which is the input channel for `writeKV`.

pump/storage/storage.go
GregoryIan: It's for debug, I really don't want to make it complicate. 😢

pump/storage/storage.go
GregoryIan: lol

suzaku:
LGTM

cmd/pump/pump.toml
july2993: if you change the definition, then it's not compatible with the old version, you should not use the same name of config

PR #633, opened 2019-06-12
Introduce slow mode to send leveldb write requests asynchronously

suzaku:

[Jira Issue](https://internal.pingcap.net/jira/browse/TOOL-1252)

When the downstream processing of `writeToValueLog` get blocked, the
buffered channel may get full and block incoming requests.

When the channel is full, switch to a slow mode.

In the slow mode, we don't send requests directly through the channel to
downstream.
Instead, a chaser is started to scan the written vlog from the last
unread pointer till the end, and send the scanned requests to the
channel that may still be full.

Once we return from the scanning, we conclude that the downstream
processing may have already resume by now.
So we try to exhaust any remaining unprocessed vlog and switch back to
the normal faster mode.
To exhaust the remaining unprocessed vlog, we need to first acquire the
write lock to prevent newer writes.
But we would limit this procedure to 3 seconds, and consider it failure
on timeout.

### Check List 

Tests 

 - Manual test
   I lowered the capacity of the buffered channel to be 10 and insert
`time.Sleep(10 * time.Seconds)` in `batchRequests` to simulate slow
processing.

Code changes

Side effects

   A write lock need to be acquired before writing a batch of vlog.
 - Increased code complexity

Related changes


pump/storage/chaser.go
```diff
@@ -0,0 +1,145 @@
...
+		return nil
+	})
+	log.Info("Finish scanning vlog", zap.Int("processed", count))
+	return errors.Trace(err)
+}
+
+func (sc *slowChaser) catchUpWithTimeout(timeout time.Duration) error {
+	errTimeout := errors.New("Recovery Timeout")
+	t0 := time.Now()
+	err := sc.vlog.scanRequests(*sc.lastUnreadPtr, func(req *request) error {
```
GregoryIan:
```
// currently we only use this in NewAppend** to scan the record which not write to KV but in the value log, so it's OK to hold the vlog.filesLock lock
func (vlog *valueLog) scan(start valuePointer, fn func(vp valuePointer, record *Record) error) error {
```

suzaku: Solved through #650

pump/storage/storage.go
```diff
@@ -847,15 +855,24 @@ func (a *Append) batchRequest(reqs chan *request, maxBatchNum int) chan []*reque
...
+			close(done)
+			cancel()
+		}()
 
 		var bufReqs []*request
 		var size int
 
 		write := func(batch []*request) {
+			slowChaser.WriteLock.Lock()
+			defer slowChaser.WriteLock.Unlock()
```
GregoryIan: I suggest don't use defer in hotpath

suzaku: `defer` is used here because:

1. `write` is a function more than 30 lines long with at least 4 exit paths, it's error-prone to insert `Unblock`s to the correct places;
1. The overhead introduced by using `defer` is trivial compared to the time it takes to save the binlog and then call `fsync`

suzaku:
Unnecessary attempts to recover from slow mode would slow down `WriteBinlog`.
To solve this problem, we can measure the time taken to catch up, and only try to recover if it's less than 1 second(or some other reasonable threshold).
If it's short enough, we may guess that:

1. There are not many logFiles scanned since the `lastUnreadPtr`
1. The output channel is not blocking

```diff
diff --git a/pump/storage/chaser.go b/pump/storage/chaser.go
index dc2c5163..aabde8f7 100644
--- a/pump/storage/chaser.go
+++ b/pump/storage/chaser.go
@@ -85,12 +85,19 @@ func (sc *slowChaser) Run(ctx context.Context) {
 			continue
 		}
 
+		t0 := time.Now()
 		err := sc.catchUp()
 		if err != nil {
 			log.Error("Failed to catch up", zap.Error(err))
 			continue
 		}
+		tCatchUp := time.Since(t0)
 
+		// Only try to recover if the catch up step takes less than 1 second,
+		// which means the slow chaser is quite close to the head of logs
+		if tCatchUp > time.Second {
+			continue
+		}
 		// Try to recover from slow mode in a limited time
 		// Once we hold the write lock, we can be sure the vlog is not being appended
 		sc.WriteLock.Lock()
diff --git a/pump/storage/storage.go b/pump/storage/storage.go
index 7a81ae02..72c34c0a 100644
--- a/pump/storage/storage.go
+++ b/pump/storage/storage.go
@@ -857,7 +857,7 @@ func (a *Append) batchRequest(reqs chan *request, maxBatchNum int) chan []*reque
 func (a *Append) writeToValueLog(reqs chan *request) chan *request {
 	ctx, cancel := context.WithCancel(context.Background())
 	done := make(chan *request, a.options.KVChanCapacity)
-	slowChaser := newSlowChaser(a.vlog, 3*time.Second, done)
+	slowChaser := newSlowChaser(a.vlog, time.Second, done)
 	go slowChaser.Run(ctx)
 
 	go func() {
```

pump/storage/storage.go
GregoryIan: ok, convincing explanation

GregoryIan:
@suzaku not sure if it is too strict, but can we also check the channel(toKV) remain length? I think it only need to satisfy one of them, then try final catch-up

GregoryIan:
I don't think there is a perfect strategy, as long as we control the number of attempts, like one time per minute

suzaku:
Below is a log snippet demonstrating how the above adjustment affect the slow chaser:

```
2019/07/03 12:02:25.292 +08:00] [INFO] [storage.go:127] [NewAppendWithResolver] [options="{\"ValueLogFileSize\":9000,\"Sync\":true,\"KVChanCapacity\":10,\"SlowWriteThreshold\":1,\"StopWriteAtAvailableSpace\":0,\"KVConfig\":null}"]
[2019/07/03 12:02:25.293 +08:00] [INFO] [storage.go:1296] ["open metadata db"] [config="{\"block-cache-capacity\":8388608,\"block-restart-interval\":16,\"block-size\":4096,\"compaction-L0-trigger\":8,\"compaction-table-size\":67108864,\"compaction-total-size\":536870912,\"compaction-total-size-multiplier\":8,\"write-buffer\":67108864,\"write-L0-pause-trigger\":24,\"write-L0-slowdown-trigger\":17}"]
[2019/07/03 12:02:25.337 +08:00] [INFO] [storage.go:210] ["Append info"] [gcTS=0] [maxCommitTS=0] [headPointer="{\"Fid\":0,\"Offset\":0}"] [handlePointer="{\"Fid\":0,\"Offset\":0}"]
[2019/07/03 12:02:25.350 +08:00] [INFO] [chaser.go:53] ["Slow chaser turned on"]
[2019/07/03 12:02:25.843 +08:00] [INFO] [chaser.go:118] ["Scanning requests to catch up with vlog"] [start="{\"Fid\":0,\"Offset\":240}"]
[2019/07/03 12:02:29.181 +08:00] [INFO] [chaser.go:126] ["Finish scanning vlog"] [processed=2522]
[2019/07/03 12:02:29.181 +08:00] [INFO] [chaser.go:118] ["Scanning requests to catch up with vlog"] [start="{\"Fid\":6,\"Offset\":9000}"]
[2019/07/03 12:02:29.184 +08:00] [INFO] [chaser.go:126] ["Finish scanning vlog"] [processed=469]
[2019/07/03 12:02:29.184 +08:00] [INFO] [chaser.go:101] ["Stopped writing temporarily to recover from slow mode"]
[2019/07/03 12:02:29.184 +08:00] [INFO] [chaser.go:131] ["Scanning requests to recover"] [start="{\"Fid\":8,\"Offset\":2650}"]
[2019/07/03 12:02:29.184 +08:00] [INFO] [chaser.go:59] ["Slow chaser turned off"]
[2019/07/03 12:02:29.184 +08:00] [INFO] [chaser.go:113] ["Successfully recover from slow mode"]
```
Note that the chaser didn't attempt to recover after the first catch up finished because it took more than 1 second.

pump/storage/chaser.go
```diff
@@ -0,0 +1,155 @@
...
+		err := sc.catchUp()
+		if err != nil {
+			log.Error("Failed to catch up", zap.Error(err))
+			continue
+		}
+		isSlowCatchUp := time.Since(t0) >= slowCatchUpThreshold
+		hasRecentRecoverAttempt := time.Since(sc.lastRecoverAttempt) <= recoveryCoolDown
+
+		if isSlowCatchUp || hasRecentRecoverAttempt {
+			continue
```
GregoryIan: output some logs here?

suzaku: Log added.

suzaku:
Some metrics I can think of that may be helpful in this scenario:

1. Count the events the slow chaser is turned on/off
1. Count failed catchUp/recovery

@GregoryIan Any suggestions?

pump/storage/chaser.go
```diff
@@ -0,0 +1,172 @@
...
+		case <-time.After(timeLeft):
+			// Return a custom error here to stop scanning
+			return errTimeout
+		}
+		return nil
+	})
+	return err
+}
+
+func (sc *slowChaser) waitUntilTurnedOn(ctx context.Context, checkInterval time.Duration) bool {
```
july2993: @return
true: it's cancel
false: we successfully wait for untilTurnedOn?

suzaku: Yes, should I make it a named returned value?

pump/storage/chaser.go
```diff
@@ -0,0 +1,172 @@
...
+	output             chan *request
+	WriteLock          sync.Mutex
+}
+
+func newSlowChaser(vlog *valueLog, recoveryTimeout time.Duration, output chan *request) *slowChaser {
+	return &slowChaser{
+		recoveryTimeout: recoveryTimeout,
+		vlog:            vlog,
+		output:          output,
+		WriteLock:       sync.Mutex{},
```
july2993: why not just sync.Mutex?

pump/storage/chaser.go
```diff
@@ -0,0 +1,172 @@
...
+		}
+
+		t0 := time.Now()
+		err := sc.catchUp()
+		if err != nil {
+			log.Error("Failed to catch up", zap.Error(err))
+			continue
+		}
+		tCatchUp := time.Since(t0)
+		slowChaserCatchUpTimeHistogram.Observe(float64(tCatchUp) / float64(time.Second))
```
july2993: you can use tCatchUp.Seconds()

pump/storage/storage.go
```diff
@@ -847,15 +855,24 @@ func (a *Append) batchRequest(reqs chan *request, maxBatchNum int) chan []*reque
...
 func (a *Append) writeToValueLog(reqs chan *request) chan *request {
+	ctx, cancel := context.WithCancel(context.Background())
 	done := make(chan *request, a.options.KVChanCapacity)
+	slowChaser := newSlowChaser(a.vlog, time.Second, done)
+	go slowChaser.Run(ctx)
 
 	go func() {
-		defer close(done)
+		defer func() {
+			close(done)
```
july2993: will put into a closed chan `done` and panic in `slowChaser` because the chaser is running

pump/storage/chaser.go
july2993: it's hard to know the behavior of `waitUntilTurnedOn` if I don't read the code, actually, I guess it will return true because it's name is `waitUntilTurnedOn` but the result is revert.

GregoryIan:
LGTM

pump/storage/storage.go
```diff
@@ -879,7 +896,19 @@ func (a *Append) writeToValueLog(reqs chan *request) chan *request {
...
+			}
+
+			if slowChaser.IsOn() {
+				return
+			}
+		SEND:
+			for _, req := range batch {
+				select {
+				case done <- req:
+				case <-time.After(10 * time.Millisecond):
```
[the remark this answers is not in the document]
suzaku: How long should we wait?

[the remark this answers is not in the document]
july2993: 3 second?

july2993:
please fix ci first

suzaku:
@july2993 I've added some new test cases, PTAL.

PR #634, opened 2019-06-13
Add config advertise-addr for drainer

### LINKED ISSUE #265: drainer should add advertise-address config
Like pump, drainer should have advertise-addr as well

suzaku:

Fix #265

Add advertise-addr config for drainer, the default value is compatible
with older version, which is the same as listen-addr.

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

Side effects

Related changes


drainer/config.go
```diff
@@ -349,3 +347,22 @@ func (cfg *Config) adjustConfig() error {
 
 	return nil
 }
+
+func validateAddr(addr string, strict bool) error {
+	urllis, err := url.Parse(addr)
+	if err != nil {
+		return errors.Errorf("failed to parse addr %v: %v", addr, err)
```
kennytm: Use `errors.Annotatef` here?

drainer/config.go
```diff
@@ -349,3 +347,22 @@ func (cfg *Config) adjustConfig() error {
...
+	}
+
+	var host string
+	if host, _, err = net.SplitHostPort(urllis.Host); err != nil {
+		return errors.Errorf("invalid host %v: %v", urllis.Host, err)
```
kennytm: ditto

drainer/config.go
suzaku: updated

drainer/config.go
```diff
@@ -349,3 +347,22 @@ func (cfg *Config) adjustConfig() error {
 
 	return nil
 }
+
+func validateAddr(addr string, strict bool) error {
+	urllis, err := url.Parse(addr)
+	if err != nil {
+		return errors.Annotate(err, fmt.Sprintf("failed to parse addr %v", addr))
```
kennytm: `Annotatef`, not `Annotate` with `Sprintf`...

suzaku: Fixed.

suzaku:
> Please add a test showing that, after calling `.adjustConfig()`, an empty `AdvertiseAddr` will be filled in with `Addr`, and an existing AdvertiseAddr will not get overridden.
> 
> Rest LGTM.

It seems like there's already one such test for `util.AdjustString` here: https://github.com/pingcap/tidb-binlog/blob/master/pkg/util/util_test.go#L207

drainer/config.go
```diff
@@ -349,3 +347,22 @@ func (cfg *Config) adjustConfig() error {
...
+		return errors.Annotatef(err, "failed to parse addr %v", addr)
+	}
+
+	var host string
+	if host, _, err = net.SplitHostPort(urllis.Host); err != nil {
+		return errors.Annotatef(err, "invalid host %v", urllis.Host)
+	}
+
+	if !util.IsValidateListenHost(host) {
+		if !(strict && host == "127.0.0.1") {
```
GregoryIan: if not strict mode , it would return error

if strict mode and host is `localhost` or `127.xx.xx.xx(not 127.0.0.1)`, it would return error

is this what your want? @suzaku

drainer/config.go
```diff
@@ -225,18 +227,12 @@ func (cfg *Config) configFromFile(path string) error {
...
 	// check ListenAddr
-	urllis, err := url.Parse(cfg.ListenAddr)
-	if err != nil {
-		return errors.Errorf("parse ListenAddr error: %s, %v", cfg.ListenAddr, err)
-	}
-
-	var host string
-	if host, _, err = net.SplitHostPort(urllis.Host); err != nil {
-		return errors.Errorf("bad ListenAddr host format: %s, %v", urllis.Host, err)
+	strictCheck := os.Getenv("BINLOG_TEST") != "1"
```
GregoryIan: put it into util pkg

drainer/config.go
suzaku:
```suggestion
		if !(strict || host == "127.0.0.1") {
```

GregoryIan:
LGTM

kennytm:
> It seems like there's already one such test for util.AdjustString here: https://github.com/pingcap/tidb-binlog/blob/master/pkg/util/util_test.go#L207

I mean specifically a test for AdvertiseAddr and Addr, as AdjustString working correctly isn't a proof that it's used correctly.

suzaku:
> > It seems like there's already one such test for util.AdjustString here: https://github.com/pingcap/tidb-binlog/blob/master/pkg/util/util_test.go#L207
> 
> I mean specifically a test for AdvertiseAddr and Addr, as AdjustString working correctly isn't a proof that it's used correctly.

OK, new cases added in `TestAdjustConfig`.

PR #635, opened 2019-06-13
tests: Reduce binlog_big test txn data

july2993:
easy to meet "TiKV Server timeout" at CI while commit this txn

easy to meet "TiKV Server timeout" at CI while commit this txn

reduce txn size of test case binlog_big

### Check List 

Tests 

 - Integration test

Code changes

Side effects

Related changes


july2993:
> LGTM, but not _big_ anymore?

not so big

PR #636, opened 2019-06-14
pkg/binlogfile: Fix test

suzaku:

One of the unit tests of pkg/binlogfile sometimes
[fails](https://internal.pingcap.net/idc-jenkins/blue/organizations/jenkins/binlog_ghpr_unit_test/detail/binlog_ghpr_unit_test/476/pipeline/):

```
[2019/06/14 09:57:29.008 +08:00] [WARN] [binlogger.go:339] ["binlog file
is already reach the gc time, but data is not send to kafka, position is
%v"]
[name=/tmp/check-6129484611666145821/2/binlog-0000000000000000-20190614095728]
[position={}]


----------------------------------------------------------------------

FAIL: binlogger_test.go:205: testBinloggerSuite.TestGC


binlogger_test.go:225:

    c.Assert(names[0], Equals, BinlogName(0))

... obtained string = "binlog-0000000000000000-20190614095728"

... expected string = "binlog-0000000000000000-20190614095729"
```

Don't use `BinlogName` when checking the result, because it calls
`time.Now` under the hood and may return different names for the same
index.

### Check List 

Tests 

 - Unit test

Code changes

Side effects

Related changes

### DIFF
--- a/pkg/binlogfile/binlogger.go
+++ b/pkg/binlogfile/binlogger.go
@@ -338,7 +338,7 @@ func (b *binlogger) GC(days time.Duration, pos binlog.Pos) {
 		} else if time.Since(fi.ModTime()) > days {
 			log.Warn(
 				"binlog file is old enough to be garbage collected, but the position is behind the safe point",
-				zap.String("name", fileName), zap.Any("position", pos))
+				zap.String("name", fileName), zap.Stringer("position", &pos))
 		}
 	}
 }
--- a/pkg/binlogfile/binlogger.go
+++ b/pkg/binlogfile/binlogger.go
@@ -336,7 +336,9 @@ func (b *binlogger) GC(days time.Duration, pos binlog.Pos) {
 			}
 			log.Info("GC binlog file", zap.String("file name", fileName))
 		} else if time.Since(fi.ModTime()) > days {
-			log.Warn("binlog file is already reach the gc time, but data is not send to kafka, position is %v", zap.String("name", fileName), zap.Reflect("position", pos))
+			log.Warn(
+				"binlog file is old enough to be garbage collected, but the position is behind the safe point",
+				zap.String("name", fileName), zap.Any("position", pos))
 		}
 	}
 }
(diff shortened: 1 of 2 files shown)


pkg/binlogfile/binlogger.go
```diff
@@ -336,7 +336,9 @@ func (b *binlogger) GC(days time.Duration, pos binlog.Pos) {
 			}
 			log.Info("GC binlog file", zap.String("file name", fileName))
 		} else if time.Since(fi.ModTime()) > days {
-			log.Warn("binlog file is already reach the gc time, but data is not send to kafka, position is %v", zap.String("name", fileName), zap.Reflect("position", pos))
+			log.Warn(
+				"binlog file is old enough to be garbage collected, but the position is behind the safe point",
+				zap.String("name", fileName), zap.Any("position", pos))
```
WangXiangUSTC: seems that `Reflect` dosen't print the pos info, do you test `Any`? and Protobuf generated `String()` function for `Pos`, maybe you `zap.Stringer` is better.

pkg/binlogfile/binlogger_test.go
```diff
@@ -214,16 +215,31 @@ func (s *testBinloggerSuite) TestGC(c *C) {
...
 	c.Assert(names, HasLen, 2)
-	c.Assert(names[0], Equals, BinlogName(0))
-	c.Assert(names[1], Equals, BinlogName(1))
+	for i, name := range names {
+		suffix, _, err := ParseBinlogName(name)
+		c.Assert(err, IsNil)
+		c.Assert(suffix, Equals, uint64(i))
+	}
+
+	// The one with index 0 should be collected
```
WangXiangUSTC: do you means the one with index 0 should be gc?

suzaku: Fixed.

pkg/binlogfile/binlogger.go
suzaku: It would be empty for `Stringer`, `Reflect` and `Any`, maybe because the `pos` we passed to this function is not initialized.

When I try logging with non empty a `pos`, I could see the value.

PR #640, opened 2019-06-19
vlog.go: Fix wrongly maintain the offset of file in memory

july2993:
Fix wrongly maintain the offset of file in memory

When the write file but may successfully write some part of data
but we don's add the offset that we maintain, later successfully
write request will get a wrong offset.

~~Add a test mimic test case by mount a specify space tmpfs, but
currently only can run in linux platform(no tmpfs in mac), and
need root privilege.~~

Add a test when no disk space fail to write and should recover after disk space are free up
set file size resource limit to make it write fail like no disk space

### Check List 

Tests 

 - Unit test
Code changes

Side effects

Related changes


pump/storage/vlog.go
```diff
@@ -329,6 +329,8 @@ func (vlog *valueLog) write(reqs []*request) error {
 
 	toDisk := func() error {
 		n, err := curFile.fd.Write(vlog.buf.Bytes())
+		atomic.AddInt64(&vlog.writableLogOffset, int64(n))
```
csuzhangxc: Can this `part of data` be ignored automatically?

pump/storage/vlog_linux_test.go
```diff
@@ -0,0 +1,86 @@
...
+	"io/ioutil"
+	"os"
+	"path"
+	"strings"
+
+	"github.com/docker/docker/pkg/mount"
+	"github.com/pingcap/check"
+)
+
+// Need root privilege for using mount to get a specify tmpfs file system
```
kennytm: IMO this is very heavy for needing `root` to run this test, and even worse binlog now depends on the entire Docker repository 🙃.

I suggest we simply set `RLIMIT_FSIZE` to a small number to ensure file size cannot increase more than 40 KB.

```go
package main

import (
	"syscall"
	"fmt"
	"os"
)

func main() {
	var origRlimit syscall.Rlimit
	err := syscall.Getrlimit(syscall.RLIMIT_FSIZE, &origRlimit)
	if err != nil { panic(err) }

	err = syscall.Setrlimit(syscall.RLIMIT_FSIZE, &syscall.Rlimit{Cur: 40, Max: origRlimit.Max})
	if err != nil { panic(err) }

	f, err := os.Create("test-fsize-rlimit.txt")
	if err != nil { panic(err) }

	defer f.Close()

	for i := 0; i < 5; i++ {
		n, err := f.Write([]byte("test content\n"))
		fmt.Println(n, err)
	}

	err = syscall.Setrlimit(syscall.RLIMIT_FSIZE, &origRlimit)
	if err != nil { panic(err) }

	for i := 0; i < 5; i++ {
		n, err := f.Write([]byte("test content\n"))
		fmt.Println(n, err)
	}
}
```

july2993: good idea
change to use `RLIMIT_FSIZE ` https://github.com/pingcap/tidb-binlog/pull/640/commits/6683f2c68f9eada5e48e18447c58f73ab925f8fc

pump/storage/vlog.go
july2993: It will skip such `part of data` if we scan the file

pump/storage/vlog.go
```diff
@@ -339,7 +341,6 @@ func (vlog *valueLog) write(reqs []*request) error {
 			}
 		}
```
suzaku: Should we make sure `fsync` is called whenever `n > 0`?

july2993: for `vlog.sync = true` case we always make sure call `fsync` before return success, currently, we will return failed if `Write` failed,  so it' s safe to not call `fsync`(we don't return success).
but note it way writes some complete record now but returns failed because we batch multi records in the `vlog.buf.Bytes`.

PR #643, opened 2019-06-20
binlogctl: return error when cmd not supported

WangXiangUSTC:
binlogctl not return error when cmd is not supported

### Check List 

Tests 

 - Manual test

Related changes

### DIFF
--- a/cmd/binlogctl/main.go
+++ b/cmd/binlogctl/main.go
@@ -17,6 +17,7 @@ import (
 	"flag"
 	"os"
 
+	"github.com/pingcap/errors"
 	"github.com/pingcap/log"
 	ctl "github.com/pingcap/tidb-binlog/binlogctl"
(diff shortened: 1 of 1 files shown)


cmd/binlogctl/main.go
```diff
@@ -59,6 +60,8 @@ func main() {
 		err = ctl.ApplyAction(cfg.EtcdURLs, node.PumpNode, cfg.NodeID, close)
 	case ctl.OfflineDrainer:
 		err = ctl.ApplyAction(cfg.EtcdURLs, node.DrainerNode, cfg.NodeID, close)
+	default:
+		err = errors.Errorf("cmd %s not supported", cfg.Command)
```
amyangfei:
```suggestion
		err = errors.NotSupportedf("cmd %s", cfg.Command)
```
use this method is simpler

PR #646, opened 2019-06-21
refine gc strategy of pump

GregoryIan:
most users forget to mark drainer offline using `binlogctl` after they stop drainer, It would block gc of pump, and then the pump will easily fill up the disk capacity because there are too many binlogs.

Regardless of ease of use or design, this design is unreasonable. There are two main reasons:
* when the disk space of the pump is full, the pump will be unavailable, which will have a great impact on the data and TiDB instance.
* drainer is a relatively separate tool and should not have too much impact on core system formation

* pump only keeps binlog events whose ts <= current time - gc(days)
* issue a critical metrics while some drainer checkpoint is old than gc ts of pump，it means binlog that drainer need is purged

### Check List 

Tests 

 - Unit test

Related changes

### DIFF
--- a/pump/metrics.go
+++ b/pump/metrics.go
@@ -42,7 +42,7 @@ var (
 		prometheus.CounterOpts{
 			Namespace: "binlog",
 			Subsystem: "pump",
-			Name:      "binlog_purge_count",
+			Name:      "detected_drainer_binlog_purge_count",
 			Help:      "binlog purge count > 0 means some unread binlog was purged",
 		}, []string{"id"})
 )
--- a/pump/metrics.go
+++ b/pump/metrics.go
@@ -37,6 +37,14 @@ var (
 			Name:      "loss_binlog_count",
 			Help:      "Total loss binlog count",
 		})
+
+	binlogPurgedCounter = prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Namespace: "binlog",
+			Subsystem: "pump",
+			Name:      "binlog_purge_count",
+			Help:      "binlog purge count > 0 means some binlog was purged",
+		}, []string{"drainer"})
 )
 
 var registry = prometheus.NewRegistry()
--- a/pump/metrics.go
+++ b/pump/metrics.go
@@ -38,13 +38,13 @@ var (
 			Help:      "Total loss binlog count",
 		})
 
-	binlogPurgedCounter = prometheus.NewCounterVec(
+	detectedDrainerBinlogPurged = prometheus.NewCounterVec(
 		prometheus.CounterOpts{
 			Namespace: "binlog",
 			Subsystem: "pump",
 			Name:      "binlog_purge_count",
 			Help:      "binlog purge count > 0 means some unread binlog was purged",
-		}, []string{"drainer"})
+		}, []string{"id"})
 )
 
 var registry = prometheus.NewRegistry()
--- a/pump/metrics.go
+++ b/pump/metrics.go
@@ -43,7 +43,7 @@ var (
 			Namespace: "binlog",
 			Subsystem: "pump",
 			Name:      "binlog_purge_count",
-			Help:      "binlog purge count > 0 means some binlog was purged",
+			Help:      "binlog purge count > 0 means some unread binlog was purged",
 		}, []string{"drainer"})
 )
 
--- a/pump/server.go
+++ b/pump/server.go
@@ -515,7 +515,7 @@ func (s *Server) gcBinlogFile() {
 			millisecond := time.Now().Add(-s.gcDuration).UnixNano() / 1000 / 1000
 			gcTS := int64(oracle.EncodeTSO(millisecond))
 
-			alertGCMS := millisecond - earlyAlertGC.Nanoseconds()/1000/1000
+			alertGCMS := millisecond + earlyAlertGC.Nanoseconds()/1000/1000
 			alertGCTS := int64(oracle.EncodeTSO(alertGCMS))
 
 			realGCTS := s.storage.GCTS(gcTS)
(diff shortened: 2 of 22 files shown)


GregoryIan:
now we check it and only output a error log, we should change the pull binlog request in other pr, and let drainer know the binlogs it request may be purged, then drainer can choose how to handle it

pump/server.go
```diff
@@ -266,6 +268,11 @@ func (s *Server) PullBinlogs(in *binlog.PullBinlogReq, stream binlog.Pump_PullBi
 	// don't use pos.Suffix now, use offset like last commitTS
 	last := in.StartFrom.Offset
 
+	gcTS := atomic.LoadInt64(&s.gcTS)
+	if last < gcTS {
+		log.Error("drainer request a purged binlog TS, some binlog events may be loss", zap.Int64("gc TS", gcTS), zap.Reflect("request", in))
+	}
```
suzaku: `Append` already has a `gcTS` field that keeps track of the same information, it seems it would be much simpler to do this check in `Append.PullCommitBinlog`.

pump/server.go
```diff
@@ -505,30 +512,30 @@ func (s *Server) gcBinlogFile() {
...
+			atomic.StoreInt64(&s.gcTS, realGCTS)
+			// detect where drainer's checkpoint is old than pump gc pointer
+			// it also means binlog events is loss for this drainer
+			s.getSafeGCTSOForDrainers(s.ctx, alertGCTS)
 		}
 	}
 }
 
-func (s *Server) getSafeGCTSOForDrainers(ctx context.Context) (int64, error) {
+func (s *Server) getSafeGCTSOForDrainers(ctx context.Context, gcTS int64) (int64, error) {
```
suzaku: I'm afraid it's not a good idea to reuse `getSafeGCTSOForDrainers` in this way. It does something that the original name doesn't imply and it changes the interface.

I think it would be more clear if we just extract another function like `getOnlineDrainers` and then reuse that function.

pump/server.go
GregoryIan: no, I want to do it in `server`, because I want to know where is the request from, even though the request doesn't have this information yet.

pump/server.go
GregoryIan: what does `getOnlineDrainers` do? I just want a function do
1. query drainer status
2. for all drainer, and check it's max commit ts

so I may need a similar function as getSafeGCTSOForDrainers

suzaku: If we have a `getOnlineDrainers` function, it can be reused in both `getSafeGCTSOForDrainers` and the function you want here, no duplication code needed and the original simple interface of `getSafeGCTSOForDrainers` doesn't have to be changed.

pump/server.go
```diff
@@ -545,6 +551,32 @@ func (s *Server) getSafeGCTSOForDrainers(ctx context.Context) (int64, error) {
 	return minTSO, nil
 }
 
+func (s *Server) detechDrainerCheckPoints(ctx context.Context, gcTS int64) {
```
suzaku: detech -> detect

pump/server.go
suzaku: Maybe we can add a `GetLastGCTS` method to `Storage`. That way we can access it here without maintaining yet another copy of the same piece of information in `Server`.

pump/server.go
GregoryIan: This is really not necessary, because you always define it when you need it, and there are not a few lines of code at all.

pump/server.go
GregoryIan: no, its's obvious that the specific time of gc is controlled outside, and it is only implemented. now we need write like this because of our bad gc implement base on leveldb

july2993: you should handle it in Storage let the user know the the the binlog is purged(maybe change PullCommitBinlog api to return error) not just check at first request,  currently, it will have some undefined behavior if  you gc a point before the point that drainer is fetching.

pump/metrics.go
```diff
@@ -37,6 +37,14 @@ var (
 			Name:      "loss_binlog_count",
 			Help:      "Total loss binlog count",
 		})
+
+	binlogPurgedCounter = prometheus.NewCounterVec(
```
july2993: s/binlogPurgedCounter/detectedDrainerBinlogPurged   ?

pump/metrics.go
```diff
@@ -37,6 +37,14 @@ var (
...
+		prometheus.CounterOpts{
+			Namespace: "binlog",
+			Subsystem: "pump",
+			Name:      "binlog_purge_count",
+			Help:      "binlog purge count > 0 means some unread binlog was purged",
+		}, []string{"drainer"})
```
july2993: s/drainer/id/   if change the metrics name

pump/metrics.go
```diff
@@ -38,13 +38,13 @@ var (
 			Help:      "Total loss binlog count",
 		})
 
-	binlogPurgedCounter = prometheus.NewCounterVec(
+	detectedDrainerBinlogPurged = prometheus.NewCounterVec(
 		prometheus.CounterOpts{
 			Namespace: "binlog",
 			Subsystem: "pump",
 			Name:      "binlog_purge_count",
```
july2993: change name too

pump/server.go
```diff
@@ -51,9 +51,11 @@ import (
...
 
 var (
-	notifyDrainerTimeout     = time.Second * 10
-	serverInfoOutputInterval = time.Second * 10
-	gcInterval               = time.Hour
+	notifyDrainerTimeout             = time.Second * 10
+	serverInfoOutputInterval         = time.Second * 10
+	gcInterval                       = time.Hour
+	earlyAlertGC                     = 20 * time.Hour
+	detectDrainterCheckpointIntervel = 10 * time.Minute
```
suzaku: detectDrainterCheckpointIntervel -> detectDrainerCheckpointInterval

PR #649, opened 2019-06-24
pump/storage: Refactor, encapsulate writing and offset updating in logFile

suzaku:

When we need to make sure the `writableOffset` is for the file
corresponding to `maxFid`, we may write code like this:

```go
if fid == maxFid {
        // get writableOffset
}
```

To make this threadsafe, we have to introduce a lock.

Separate the concern of the struct `valueLog` and `logFile`.

`valueLog` is a higher level interface to the binlog data files, it
manages a directory of `logFile`s.

`logFile` is responsible for saving data to individual files.

By allowing `LogFile` to manage its own `writeOffset`, we can be sure
that the offset is for the "current" file without using a lock.

### Check List 

Tests 

 - Unit test

Code changes

Side effects

Related changes

### DIFF
--- a/pump/storage/log_test.go
+++ b/pump/storage/log_test.go
@@ -22,6 +22,7 @@ import (
 	"math/rand"
 	"os"
 	"path"
+	"path/filepath"
 	"strconv"
 
 	fuzz "github.com/google/gofuzz"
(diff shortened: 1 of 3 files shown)


pump/storage/log_test.go
```diff
@@ -73,6 +74,18 @@ func (lfs *LogFileSuit) TearDownTest(c *check.C) {
...
+func (lfs *LogFileSuit) TestWriteOffset(c *check.C) {
+	dir := c.MkDir()
+	f, err := newLogFile(1024, filepath.Join(dir, "1024.vlog"))
+	c.Assert(err, check.IsNil)
+	c.Assert(f.GetWriteOffset(), check.Equals, int64(0))
+
+	data := make([]byte, 379)
+	err = f.Write(data, true)
+	c.Assert(err, check.IsNil)
+	c.Assert(f.GetWriteOffset(), check.Equals, int64(379))
```
csuzhangxc: Is it a _write failed_ case (use `RLIMIT_FSIZE` as before) need?

suzaku: That case was already covered elsewhere.

GregoryIan: better to migrate the case here?

GregoryIan:
Rest LGTM

pump/storage/log_test.go
suzaku: Maybe leave that for another PR?

GregoryIan: are there something undefined and need to be handled separately in another PR?

PR #650, opened 2019-06-25
pump/storage: Only hold the lock when collecting logFiles to scan

suzaku:

The original implementation of `valueLog.scan` assumes it
would only be called when pump is starting up, during which time no
writing is happening.
So it's OK to hold a lock that blocks `valueLog.write`.

There are 2 steps in `valueLog.write` that use the `filesLock`:
1. Looking up the `logFile` for `maxFid` (Read)
1. Creating a new `logFile` (Write)

Only the second case requires a write lock.

If we make `scan` acquire the read lock instead, case 1 would not be
blocked at all.
But the second case would still be blocked if we need to hold the read
lock in scan until all logFiles are processed.

1. Use `RLock` instead to get all the selected logFiles.
1. Scan without holding the lock.

This is acceptable because:
1. If `scan` is used when pump is starting up, no concurrent writes
   exist, so it's OK to go without a lock.
1. If `scan` is used to help a slow or paused LevelDB catch up with
   valueLog like in #633, it's OK to miss logFiles created during the
   scanning, because we would keep track of the pointer and retry
   scanning until the LevelDB is uptodate.

### Check List 

Tests 

 - Unit test

Code changes

Side effects

Related changes

### DIFF
--- a/pump/storage/vlog.go
+++ b/pump/storage/vlog.go
@@ -408,18 +408,20 @@ func (vlog *valueLog) scanRequests(start valuePointer, fn func(*request) error) 
 
 // currently we only use this in NewAppend** to scan the record which not write to KV but in the value log, so it's OK to hold the vlog.filesLock lock
 func (vlog *valueLog) scan(start valuePointer, fn func(vp valuePointer, record *Record) error) error {
-	vlog.filesLock.Lock()
-	defer vlog.filesLock.Unlock()
-
+	vlog.filesLock.RLock()
 	fids := vlog.sortedFids()
-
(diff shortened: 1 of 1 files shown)


GregoryIan:
scan in a snapshot is easier to understand

suzaku:
> make clear how scan can be used, after this PR if you run `scan` at run time some files may be gc and delete in `vlog.gcTS`
> rest LGTM

Document added.

suzaku:
@GregoryIan @july2993 PTAL

GregoryIan:
LGTM

PR #651, opened 2019-06-25
pump/storage: Avoid big inputs that slow down the unit test

suzaku:

With big inputs like "writing a 500MB binlog", the unit tests of
pump/storage can be very slow.

1. Remove big inputs.
1. Batch valueLog writes

### Check List 

Tests 

 - Unit test

Code changes

Side effects

Related changes

### DIFF
--- a/pump/storage/storage_test.go
+++ b/pump/storage/storage_test.go
@@ -247,6 +247,8 @@ func (as *AppendSuit) TestReadWriteGCTS(c *check.C) {
 }
 
 // test helper to write binlogNum binlog to append
+// If `prewriteValueSize` is less than or equal to 0, the size of prewriteValue for
+// binlogs will be random integers between [1, 1024].
 func populateBinlog(b Log, append *Append, prewriteValueSize int, binlogNum int32) {
 	fixedValueSize := prewriteValueSize > 0
 	var prewriteValue []byte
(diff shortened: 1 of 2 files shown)


pump/storage/storage_test.go
```diff
@@ -281,10 +276,15 @@ func populateBinlog(b Log, append *Append, prewriteValueSize int, binlogNum int3
 				binlog.Tp = pb.BinlogType_Prewrite
 				startTS := getTS()
 				binlog.StartTs = startTS
-				binlog.PrewriteValue = prewriteValue
+				if fixedValueSize {
+					binlog.PrewriteValue = prewriteValue
+				} else {
+					size := 1 + rand.Intn(len(prewriteValue)-1)
```
july2993: you want `1 + rand.Intn(len(prewriteValue)` ?

pump/storage/storage_test.go
```diff
@@ -257,17 +248,21 @@ func (as *AppendSuit) TestReadWriteGCTS(c *check.C) {
 
 // test helper to write binlogNum binlog to append
 func populateBinlog(b Log, append *Append, prewriteValueSize int, binlogNum int32) {
```
july2993: add comment on how  's behavior  when `prewriteValueSize`  <= 0,  and seems no where use it as `prewriteValueSize > 0 ` now for `fixed stype` now?

pump/storage/storage_test.go
suzaku: It should be 1 less than the size because we want it to be `1 + [0,n-2] = [1, n-1]`.

PR #656, opened 2019-06-29
bestEffort: make `GC` and `pull binlog` behavior more accurate 

GregoryIan:
* GC would delete the binlogs still not sorted in vlog
* pull binlog  request would miss some purged binlog while reading and GC are happening simultaneously in the same range

* store  ts of purged binlog  as gc TS, it's more accurate
* prevent to delete not sorted binlog
* return `ErrReadPurgedBinlog` while find binlog meta in kv, but not found its data in vlog
* check the request `last ts` wether less than `gcTS`. if `gcTS` is large than `last ts`, it means some binlogs  in [last ts, gc ts] were purged

but still some condition we can't handled, like
* we don't find matched p-binlog of c-binlog in leveldb, it may be caused
   * not write p-binlog
   * p-binlog was purged
we really only need to store the information we need in leveldb, but we don't do it like it now, it's unresolved now

### Check List 

Tests 

 - Unit test
 - Integration test


pump/storage/storage.go
```diff
@@ -1027,6 +1055,14 @@ func (a *Append) PullCommitBinlog(ctx context.Context, last int64) <-chan []byte
 			irange.Limit = encodeTSKey(atomic.LoadInt64(&a.maxCommitTS) + 1)
 			iter := a.metadata.NewIterator(irange, nil)
 
+			gcTS := atomic.LoadInt64(&a.gcTS)
+			if last < gcTS {
+				log.Warn("some binlogs was pugred, cause read from last ts is less than gcTS", zap.Int64("last ts", last), zap.Int64("gcTS", gcTS))
+				iter.Release()
+				a.responseBinlogError(ctx, responseChan, errors.Annotatef(butil.ErrorPurgedBinlog, "read binlog from last ts %d", last))
+				return
+			}
```
suzaku: Can we do this outside the `for` ?

pump/storage/storage.go
```diff
@@ -666,14 +678,20 @@ func (a *Append) doGCTS(ts int64) {
...
-		var lastKey []byte
 
 		for iter.Next() && deleteBatch < 100 {
 			batch.Delete(iter.Key())
 			deleteNum++
-			lastKey = iter.Key()
+			cuurentTS := decodeTSKey(iter.Key())
+
+			// don't delete the ts not sorted
+			if cuurentTS >= atomic.LoadInt64(&a.maxCommitTS) {
```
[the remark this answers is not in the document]
GregoryIan: sure

pump/storage/storage.go
```diff
@@ -976,7 +997,7 @@ func (a *Append) feedPreWriteValue(cbinlog *pb.Binlog) error {
 
 	pvalue, err := a.vlog.readValue(vp)
 	if err != nil {
-		return errors.Annotatef(err, "read P-Binlog value failed, vp: %+v", vp)
+		return errors.Annotatef(butil.ErrorPurgedBinlog, "read P-Binlog value failed, vp: %+v", vp)
```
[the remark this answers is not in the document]
GregoryIan: leave it to you, just return err now

pump/storage/storage.go
GregoryIan: not for it, but can for the response error below

GregoryIan:
@suzaku PTAL

pkg/util/util.go
```diff
@@ -30,6 +30,11 @@ import (
 	"go.uber.org/zap"
 )
 
+var (
+	// ErrorPurgedBinlog indicates binlog was purged
+	ErrorPurgedBinlog = errors.New("binlog was purged")
```
suzaku: This error should belong to the `pump` package?

GregoryIan: no need, maybe other package need to access it

july2993:
if you want to change to store ts of purged binlog as gc TS, no need `orward a little bit to make sure we can get the according P binlog` in `append.GC()` now

GregoryIan:
I don't understand what your meaning is @july2993, I know I can get start ts from c-binlog, then?

july2993:
> I don't understand what your meaning is @july2993, I know I can get start ts from c-binlog, then?

then you can know it's purged or not

GregoryIan:
you don't understand my thoughts. you give a strong consensus that  **c binlog and p binlog are definitely in a same pump** . but which one will  guarantee it? tidb and pump client, not pump storage. as your comment leaved before 
```
// In pump-client, a C-binlog should always be sent to the same pump instance as the matching P-binlog.
// But in some older versions of pump-client, writing of C-binlog would fallback to some other instances when the correct one is unavailable.
// When this error occurs, we may assume that the matching P-binlog is on a different pump instance.
// And it would  query TiKV for the matching C-binlog. So it should be OK to ignore the error here.
```

GregoryIan:
any problems need to be adjusted?

GregoryIan:
I will close this pr, and take easy to open when you need it

PR #661, opened 2019-07-02
pump: Add more tests for pump/storage

suzaku:

Some recently added functions are not tested.

Add a test.

### Check List 

Tests 

 - Unit test

Code changes

Side effects

Related changes

### DIFF
--- a/pump/storage/vlog.go
+++ b/pump/storage/vlog.go
@@ -317,6 +317,7 @@ func (vlog *valueLog) write(reqs []*request) error {
 	vlog.filesLock.RUnlock()
 
 	var bufReqs []*request
+	vlog.buf.Reset()
 
 	toDisk := func() error {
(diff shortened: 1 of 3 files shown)


suzaku:
*NOTE*: Please don't retry the integration tests, I'm investigating the test case.

pump/storage/vlog.go
```diff
@@ -317,6 +317,7 @@ func (vlog *valueLog) write(reqs []*request) error {
 	vlog.filesLock.RUnlock()
 
 	var bufReqs []*request
+	vlog.buf.Reset()
```
csuzhangxc: `Reset` when entering this function VS when leaving this function, which is better?

suzaku: It makes little difference in our set up. Since my intention is to never use any data left over in the buffer, I think it's clearer to reset the buffer upfront.

suzaku:
@csuzhangxc PTAL

PR #663, opened 2019-07-02
refine gc strategy of pump (#646)

GregoryIan:
most users forget to mark drainer offline using `binlogctl` after they stop drainer, It would block gc of pump, and then the pump will easily fill up the disk capacity because there are too many binlogs.

Regardless of ease of use or design, this design is unreasonable. There are two main reasons:
* when the disk space of the pump is full, the pump will be unavailable, which will have a great impact on the data and TiDB instance.
* drainer is a relatively separate tool and should not have too much impact on core system formation

* pump only keeps binlog events whose ts <= current time - gc(days)
* issue a critical metrics while some drainer checkpoint is old than gc ts of pump，it means binlog that drainer need is purged

### Check List 

Tests 

 - Unit test

Related changes


mahjonp:
/run-all-tests tidb=release-2.1 tikv=release-2.1 pd=release-2.1

suzaku:
/run-all-tests tidb=release-2.1 tikv=release-2.1 pd=release-2.1

GregoryIan:
@july2993 PTAL

PR #666, opened 2019-07-04
pump: Make the pump/storage unit tests faster

suzaku:

It still takes about 15 seconds to run.

Make it faster, the new run time is about 6.5 seconds.

### Check List 

Tests 

 - Unit test

Code changes

Side effects

Related changes

### DIFF
--- a/pump/storage/storage.go
+++ b/pump/storage/storage.go
@@ -907,7 +907,7 @@ func (a *Append) writeToValueLog(reqs chan *request) chan *request {
 					size = 0
 					bufReqs = bufReqs[:0]
 				}
-			case <-time.After(time.Millisecond):
+			default:
 				if len(bufReqs) > 0 {
 					write(bufReqs)
 					size = 0
--- a/pump/storage/storage.go
+++ b/pump/storage/storage.go
@@ -907,7 +907,7 @@ func (a *Append) writeToValueLog(reqs chan *request) chan *request {
 					size = 0
 					bufReqs = bufReqs[:0]
 				}
-			default:
+			case <-time.After(time.Millisecond):
 				if len(bufReqs) > 0 {
 					write(bufReqs)
 					size = 0
--- a/pump/storage/bench_test.go
+++ b/pump/storage/bench_test.go
@@ -75,7 +75,7 @@ func benchmarkPull(b *testing.B, prewriteValueSize int, binlogNum int) {
 	append := newAppend(b)
 	defer os.RemoveAll(append.dir)
 
-	populateBinlog(b, append, prewriteValueSize, int32(binlogNum))
+	populateBinlog(b, append, prewriteValueSize, binlogNum)
 
 	runtime.GC()
 	b.ResetTimer()
--- a/pump/storage/storage_test.go
+++ b/pump/storage/storage_test.go
@@ -263,7 +263,7 @@ func populateBinlog(b Log, append *Append, prewriteValueSize int, binlogNum int)
 
 	seq := make(chan struct{}, 10)
 	go func() {
-		for i := 0 ; i < binlogNum; i++ {
+		for i := 0; i < binlogNum; i++ {
 			seq <- struct{}{}
 		}
 		close(seq)
--- a/pump/storage/storage_test.go
+++ b/pump/storage/storage_test.go
@@ -118,7 +118,7 @@ func (as *AppendSuit) TestWriteBinlogAndPullBack(c *check.C) {
 	as.testWriteBinlogAndPullBack(c, -1, 1024)
 }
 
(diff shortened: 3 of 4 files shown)


pump/storage/storage.go
```diff
@@ -907,7 +907,7 @@ func (a *Append) writeToValueLog(reqs chan *request) chan *request {
 					size = 0
 					bufReqs = bufReqs[:0]
 				}
-			default:
+			case <-time.After(time.Millisecond):
```
july2993: I don't think this is a good idea for the latency, and what  benefit can we get?

suzaku: I thought `default` was OK until I checked how many requests got written in each bulk (the number of concurrent workers was set to 256):

```
[2019/07/05 09:33:42.458 +08:00] [INFO] [vlog.go:315] ["Writing requests in bulk"] [number=2]
[2019/07/05 09:33:42.475 +08:00] [INFO] [vlog.go:315] ["Writing requests in bulk"] [number=239]
[2019/07/05 09:33:42.492 +08:00] [INFO] [vlog.go:315] ["Writing requests in bulk"] [number=39]
[2019/07/05 09:33:42.508 +08:00] [INFO] [vlog.go:315] ["Writing requests in bulk"] [number=217]
[2019/07/05 09:33:42.530 +08:00] [INFO] [vlog.go:315] ["Writing requests in bulk"] [number=164]
[2019/07/05 09:33:42.551 +08:00] [INFO] [vlog.go:315] ["Writing requests in bulk"] [number=249]
[2019/07/05 09:33:42.568 +08:00] [INFO] [vlog.go:315] ["Writing requests in bulk"] [number=176]
[2019/07/05 09:33:42.582 +08:00] [INFO] [vlog.go:315] ["Writing requests in bulk"] [number=84]
[2019/07/05 09:33:42.600 +08:00] [INFO] [vlog.go:315] ["Writing requests in bulk"] [number=248]
[2019/07/05 09:33:42.616 +08:00] [INFO] [vlog.go:315] ["Writing requests in bulk"] [number=190]
[2019/07/05 09:33:42.632 +08:00] [INFO] [vlog.go:315] ["Writing requests in bulk"] [number=255]
[2019/07/05 09:33:42.647 +08:00] [INFO] [vlog.go:315] ["Writing requests in bulk"] [number=77]
[2019/07/05 09:33:42.660 +08:00] [INFO] [vlog.go:315] ["Writing requests in bulk"] [number=108]
```

As you can see, the bulk could be really small, eg. 2.

If we try to set a short timeout like (1ms or 500us), more requests can be written together:

```
[2019/07/05 09:35:11.199 +08:00] [INFO] [vlog.go:315] ["Writing requests in bulk"] [number=256]
[2019/07/05 09:35:11.211 +08:00] [INFO] [vlog.go:315] ["Writing requests in bulk"] [number=256]
[2019/07/05 09:35:11.227 +08:00] [INFO] [vlog.go:315] ["Writing requests in bulk"] [number=247]
[2019/07/05 09:35:11.239 +08:00] [INFO] [vlog.go:315] ["Writing requests in bulk"] [number=256]
[2019/07/05 09:35:11.253 +08:00] [INFO] [vlog.go:315] ["Writing requests in bulk"] [number=255]
[2019/07/05 09:35:11.264 +08:00] [INFO] [vlog.go:315] ["Writing requests in bulk"] [number=256]
[2019/07/05 09:35:11.279 +08:00] [INFO] [vlog.go:315] ["Writing requests in bulk"] [number=252]
[2019/07/05 09:35:11.291 +08:00] [INFO] [vlog.go:315] ["Writing requests in bulk"] [number=256]
[2019/07/05 09:35:11.302 +08:00] [INFO] [vlog.go:315] ["Writing requests in bulk"] [number=14]
```

july2993: it's ok for this, because the upstream has no more data, you should test if this can improve throughput but not more written together.

suzaku: It only waits when the channel is empty, not for **every** write. As you can see in the logs above, it can actually save us some `fsync` calls which take much longer than `1ms`.

july2993: it will, you can use one concurrent write and check it.

suzaku: But is that a real case? That the client only writes binlogs one by one in a single thread.

july2993: it's not about how the client write, just let's you know the problem, and this get not benefit to wait 1ms, if it run into the default case it means no more request buffer now, you can write the current one ASAP, and you can alwasy batch write the later ones if have more than one, at less you must show this change can improve throughput for sacrifice in some case.

here is one metrics about write qps and latency of the WriteBinlog RPC
Screen Shot 2019-07-05 at 11 11 18 AM

suzaku:
@july2993 PTAL

WangXiangUSTC:
LGTM

PR #667, opened 2019-07-04
drainer: fix failure to update BIT columns with downstream TiDB when message passes through Kafka

lichunzhu:
Issue TOOL-1374, which is caused by [pingcap/tidb#10988](https://github.com/pingcap/tidb/issues/10988)

Encode BIT types as integers instead of byte strings. Since the longest value is BIT(64), it is safe to always convert as uint64.
1. This PR is an improved version of [pingcap/binlog PR#655](https://github.com/pingcap/tidb-binlog/pull/655). It mainly puts test in tests/update_bit1/ into tests/dailytest/case.go. This can simplify test code and test both drainer -> mysq and drainer -> kafka -> mysql.
2. This PR also encodes BIT types as integers in pkg/loader/translate.go to avoid fault in [pingcap/tidb#10988](https://github.com/pingcap/tidb/issues/10988).
3. Change the table name in tests/dailytest/case.go from 'a' to 'binlog_case3' to make it easier for debug.

### Check List 

Tests 

    * Verified the error in TOOL-1374 is no longer reproducible.

Code changes

Side effects

Related changes


pkg/loader/translate.go
```diff
@@ -84,6 +85,10 @@ func columnToArg(mysqlType string, c *pb.Column) (arg interface{}) {
 			var str string = string(c.GetBytesValue())
 			return str
 		}
+		if mysqlType == "bit" {
+			val, _ := BinaryLiteral(c.GetBytesValue()).ToInt(nil)
```
july2993: handle err

tests/dailytest/case.go
```diff
@@ -106,18 +106,34 @@ var case2Clean = []string{`
...
-ALTER TABLE a ADD UNIQUE INDEX aidx(a1);
+ALTER TABLE binlog_case3 ADD UNIQUE INDEX aidx(a1);
 `,
 }
 
 var case3Clean = []string{
-	`DROP TABLE a`,
+	`DROP TABLE binlog_case3`,
+}
+
```
july2993: add some comment context like the origin "# Verify fix for TOOL-1346"

pkg/loader/translate.go
lichunzhu: > handle err

Already modified

tests/dailytest/case.go
lichunzhu: > add some comment context like the origin "# Verify fix for TOOL-1346"

Already added

tests/dailytest/case.go
```diff
@@ -105,6 +105,7 @@ var case2Clean = []string{`
 	DROP TABLE binlog_case2`,
 }
 
+// Table name changed in TOOL-1346
```
july2993: ?? remove this

pkg/loader/translate.go
```diff
@@ -84,6 +86,17 @@ func columnToArg(mysqlType string, c *pb.Column) (arg interface{}) {
 			var str string = string(c.GetBytesValue())
 			return str
 		}
+		// issues: TOOL-1346 and TOOL-1374
+		// for downstream = mysql
+		// it work for tidb to update binary by translating []byte to uint64
```
july2993: replace comments like this:
```
https://github.com/pingcap/tidb/issues/10988
Binary literal is not passed correctly for TiDB in some cases, so encode BIT types as integers instead of byte strings. Since the longest value is BIT(64), it is safe to always convert as uint64
```

tests/dailytest/case.go
```diff
@@ -105,19 +105,38 @@ var case2Clean = []string{`
...
 `,
 }
 
 var case3Clean = []string{
-	`DROP TABLE a`,
+	`DROP TABLE binlog_case3`,
+}
+
+// Ref issue: TOOL-1346
+// test bit value synchronization for drainer -> mysql and drainer -> kafka -> mysql
```
july2993: remove this, just ` Test issue: TOOL-1346`

pkg/loader/translate.go
```diff
@@ -84,6 +86,17 @@ func columnToArg(mysqlType string, c *pb.Column) (arg interface{}) {
...
+		if mysqlType == "bit" {
+			val, err := types.BinaryLiteral(c.GetBytesValue()).ToInt(nil)
```
july2993: change `columnToArg` to return (arg interface{}, err error)
can't just return nil, then the caller can't handle it

arbiter/server.go
```diff
@@ -266,7 +266,10 @@ func syncBinlogs(source <-chan *reader.Message, ld loader.Loader) {
 	dest := ld.Input()
 	for msg := range source {
 		log.Debug("recv msg from kafka reader", zap.Int64("ts", msg.Binlog.CommitTs), zap.Int64("offset", msg.Offset))
-		txn := loader.SlaveBinlogToTxn(msg.Binlog)
+		txn, err := loader.SlaveBinlogToTxn(msg.Binlog)
+		if err != nil {
```
july2993: do not silently, if err is not nil, txn is not a valid value, think about how to handle it

pkg/loader/translate.go
```diff
@@ -15,10 +15,11 @@ package loader
 
 import (
 	pb "github.com/pingcap/tidb-tools/tidb-binlog/slave_binlog_proto/go-binlog"
+	types "github.com/pingcap/tidb/types"
```
july2993: no need types ?

pkg/loader/translate.go
```diff
@@ -36,44 +37,53 @@ func SlaveBinlogToTxn(binlog *pb.Binlog) (txn *Txn) {
 				dml.Tp = getDMLType(mut)
 
 				// setup values
-				dml.Values = getColVals(table, mut.Row.GetColumns())
+				dml.Values, err = getColVals(table, mut.Row.GetColumns())
+				if err != nil {
+					return txn, err
```
july2993: return ni, err

pkg/loader/translate.go
```diff
@@ -36,44 +37,53 @@ func SlaveBinlogToTxn(binlog *pb.Binlog) (txn *Txn) {
...
+				}
 
 				// setup old values
 				if dml.Tp == UpdateDMLType {
-					dml.OldValues = getColVals(table, mut.ChangeRow.GetColumns())
+					dml.OldValues, err = getColVals(table, mut.ChangeRow.GetColumns())
+					if err != nil {
+						return txn, err
```
july2993: return nil, err

tests/kafka/kafka.go
```diff
@@ -89,7 +89,12 @@ func main() {
 			case msg := <-breader.Messages():
 				str := msg.Binlog.String()
 				log.S().Debugf("recv: %.2000s", str)
-				ld.Input() <- loader.SlaveBinlogToTxn(msg.Binlog)
+				txn, err := loader.SlaveBinlogToTxn(msg.Binlog)
+				if err != nil {
+					log.S().Error(errors.ErrorStack(err))
+					log.S().Fatal(err)
```
july2993: duplicated log, if you want Fatal(err) why need the previous Error log, you can just Fatale here

arbiter/server.go
```diff
@@ -262,16 +279,27 @@ func (s *Server) loadStatus() (int, error) {
...
-		txn := loader.SlaveBinlogToTxn(msg.Binlog)
+		if err != nil {
+			continue
+		}
+		txn, err = loader.SlaveBinlogToTxn(msg.Binlog)
+		if err != nil {
+			log.Error("transfer binlog failed, program will stop handling data from loader", zap.Error(err))
+			res <- err
+			continue
+		}
```
suzaku: It seems like we could just return the error here. If I didn't misunderstand the code above, after the first error, all messages would simply be skipped, so unless we had to read all the messages from the channel, we can just return.

arbiter/server_test.go
```diff
@@ -407,7 +407,18 @@ func (s *syncBinlogsSuite) TestShouldSendBinlogToLoader(c *C) {
 	}()
 	ld := dummyLoader{input: dest}
 
-	syncBinlogs(source, &ld)
+	res := make(chan error)
+	var err error = nil
```
suzaku: Note that Go initialize all variables to the empty value of its type, so it's OK to just use `var err error` here.

pkg/loader/translate.go
```diff
@@ -15,10 +15,11 @@ package loader
 
 import (
 	pb "github.com/pingcap/tidb-tools/tidb-binlog/slave_binlog_proto/go-binlog"
+	"github.com/pingcap/tidb/types"
 )
 
 // SlaveBinlogToTxn translate the Binlog format into Txn
-func SlaveBinlogToTxn(binlog *pb.Binlog) (txn *Txn) {
+func SlaveBinlogToTxn(binlog *pb.Binlog) (txn *Txn, err error) {
```
suzaku:
```suggestion
func SlaveBinlogToTxn(binlog *pb.Binlog) (*Txn, error) {
```

tests/dailytest/case.go
```diff
@@ -106,18 +106,35 @@ var case2Clean = []string{`
...
 `,
 }
 
 var case3Clean = []string{
-	`DROP TABLE a`,
+	`DROP TABLE binlog_case3`,
+}
+
+// Test issue: TOOL-1346
+var case4 = []string{`
```
suzaku: To further improve readability, you can rename the case to tell the maintainer what it's about, for example, `caseTblWithGeneratedCol` and `caseSetBitColumn` would tell us far more than just `case1, 2, 3, 4`.

arbiter/server.go
suzaku: According to @lichunzhu , the goroutine running `kafkaReader` may block waiting to send message to a full channel if we just return here.
If `kafkaReader` is closed, the the goroutine would panic because it attempts to send to a closed channel.
Anyhow, there's space of improvement in the implementation of the `reader`, so @lichunzhu  will take care to fix `reader` first.

pkg/loader/translate.go
lichunzhu: revised

tests/dailytest/case.go
lichunzhu: revised

arbiter/server.go
lichunzhu: Now this function will return immediately

lichunzhu:
All comments have been revised. PTAL. @suzaku

PR #670, opened 2019-07-08
drainer: fix ts map

WangXiangUSTC:
fix issue: https://internal.pingcap.net/jira/browse/TOOL-1296

get ts from downstream after batch is executed.

### Check List 

Tests 

 - Unit test
 - Manual test

Related changes


pkg/loader/load_test.go
```diff
@@ -473,3 +475,38 @@ func (s *runSuite) TestShouldFlushWhenInputIsEmpty(c *check.C) {
...
+	updateLastFinishTSInterval = 0
+	defer func() {
+		updateLastFinishTSInterval = originUpdateLastFinishTSInterval
+	}()
+	txns = []*Txn{
+		{Metadata: 7},
+		{Metadata: 8},
+	}
+	loader.markSuccess(txns...)
+	c.Assert(txns[len(txns)-1].FinishTS, check.Equals, int64(88881234))
```
suzaku: The original assertion is to make sure that when we call `markSuccesses` again no `FinishTS` are set, so it should be `0`.

WangXiangUSTC: updated in https://github.com/pingcap/tidb-binlog/pull/670/commits/8b1e5bde976da233ddbcdd4ef36291ef59b421ba

suzaku:
LGTM

drainer/sync/mysql.go
```diff
@@ -38,14 +38,14 @@ type MysqlSyncer struct {
...
-func NewMysqlSyncer(cfg *DBConfig, tableInfoGetter translator.TableInfoGetter, worker int, batchSize int, queryHistogramVec *prometheus.HistogramVec, sqlMode *string) (*MysqlSyncer, error) {
+func NewMysqlSyncer(cfg *DBConfig, tableInfoGetter translator.TableInfoGetter, worker int, batchSize int, queryHistogramVec *prometheus.HistogramVec, sqlMode *string, destDBType string) (*MysqlSyncer, error) {
 	db, err := createDB(cfg.User, cfg.Password, cfg.Host, cfg.Port, sqlMode)
 	if err != nil {
 		return nil, errors.Trace(err)
 	}
 
 	var opts []loader.Option
-	opts = append(opts, loader.WorkerCount(worker), loader.BatchSize(batchSize))
+	opts = append(opts, loader.WorkerCount(worker), loader.BatchSize(batchSize), loader.DownstreamTp(destDBType))
```
GregoryIan: is definition of `DownstreamTp` tricky?  you adjust type of downstream, and only want to use it to check whether we need to keep slave TS.  I suggest that  it's better not to covert it to other type, it only make some people confused

drainer/sync/syncer.go
```diff
@@ -24,6 +24,9 @@ type Item struct {
 	PrewriteValue *pb.PrewriteValue // only for DML
 	Schema        string
 	Table         string
+
+	// the finish TS executed in downstream, only for tidb
+	FinishTS int64
```
GregoryIan: is `AppliedTS` better?

drainer/syncer.go
```diff
@@ -203,9 +206,10 @@ func (s *Syncer) handleSuccess(fakeBinlog chan *pb.Binlog, lastTS *int64) {
 				atomic.StoreInt64(lastTS, ts)
 			}
 
-			// save ASAP for DDL
-			if item.Binlog.DdlJobId > 0 {
+			// save ASAP for DDL, and if FinishTS > 0, we should save the ts map
+			if item.Binlog.DdlJobId > 0 || item.FinishTS > 0 {
 				saveNow = true
+				finishTS = item.FinishTS
```
GregoryIan: where to reset `finishTS`

pkg/loader/load.go
```diff
@@ -181,6 +206,10 @@ func (s *loaderImpl) GetSafeMode() bool {
 }
 
 func (s *loaderImpl) markSuccess(txns ...*Txn) {
```
GregoryIan: does `markSuccess` take effect  synchronously or asynchronously?

WangXiangUSTC: `markSuccess` can take effect asynchronously, but if we get applied ts here, can only run synchronously after execute DML and DDL

drainer/syncer.go
WangXiangUSTC: updated https://github.com/pingcap/tidb-binlog/pull/670/commits/4c8782f0e021d79d1b1ecb9094716dc0a9810c2f

drainer/sync/syncer.go
WangXiangUSTC: updated https://github.com/pingcap/tidb-binlog/pull/670/commits/4c8782f0e021d79d1b1ecb9094716dc0a9810c2f

drainer/sync/mysql.go
WangXiangUSTC: updated https://github.com/pingcap/tidb-binlog/pull/670/commits/412a5226bcd11cac6e2422db7c47e3eecc8a6306

july2993:
use github.com/pingcap/parser/mysql instead of github.com/pingcap/tidb/mysql
@WangXiangUSTC pls fix ci

WangXiangUSTC:
@GregoryIan @july2993 PTAL

GregoryIan:
LGTM
and  @july2993 PTAL

PR #671, opened 2019-07-09
Set finish ts in markSuccesses which is a natural match to handleSuccesses

suzaku:
pkg/loader: set finish ts in markSuccesses

`batchManager` is a helper struct to handle batch execution of transactions.
What we need here is to set the matching slave ts of a transaction successfully synced to the downstream/slave, so I think `markSuccesses` is a better place.

Refactor by moving the setting finish ts implementation to `markSuccesses`.

### Check List 

Tests 

 - Unit test

Code changes

Side effects

Related changes


WangXiangUSTC:
LGTM

PR #673, opened 2019-07-09
update parser dep

july2993:
update parser to the latest version.
also, update tidb/pd/..  and many fix about API or definition change.

update parser/tidb/pd to latest version

### Check List 

Tests 

 - Unit test
 - Integration test
Code changes

Side effects

 - Increased code complexity

Related changes


pkg/util/p8s.go
```diff
@@ -50,7 +50,13 @@ func (mc MetricClient) Start(ctx context.Context, grouping map[string]string) {
...
 			return
 		case <-time.After(mc.interval):
-			if err := addToPusher("binlog", grouping, mc.addr, mc.registry); err != nil {
+			pusher := push.New(mc.addr, "binlog")
+			// add grouping
+			for k, v := range grouping {
+				pusher = pusher.Grouping(k, v)
+			}
+			pusher = pusher.Gatherer(mc.registry)
+			if err := pusher.Add(); err != nil {
```
suzaku: We don't have to remove the unit tests. Just define a function that acts like the `push.AddFromGatherer` and assign it to `addToPusher`.

suzaku:
```go
func addFromGatherer(job string, grouping map[string]string, url string, g promethesus.Gatherer) error {
			pusher := push.New(url, job)
			// add grouping
			for k, v := range grouping {
				pusher = pusher.Grouping(k, v)
			}
			pusher = pusher.Gatherer(g)
			return pusher.Add()
}
```

july2993: change in https://github.com/pingcap/tidb-binlog/pull/673/commits/ad0b432cd4ab5df343eaa2d69c6fd46bd9cf4ca7

PR #674, opened 2019-07-09
handle Recover job and fix can't close internally

july2993:
reviewed after https://github.com/pingcap/tidb-binlog/pull/673 

- handle Recover job, don't error in schema by falling to the default case inside `handleDDL`
- fix can't close internally
    Note if in the taskGroup.Go if we call s.Close()
    it will wait on the taskGroup in the Close() implement
    but we are still clock in the closure pass into Go for calling
    s.Close(), so it will call Done inside taskGroup and will block in
    Wait(), this is a dead lock.
### Check List 

Tests 

 - Unit test
 - Integration test
run `recover table a` and fail(exec at downstream and the downstream not support) and quit

Code changes

Side effects

 - Increased code complexity

Related changes

### DIFF
--- a/drainer/server.go
+++ b/drainer/server.go
@@ -216,7 +216,7 @@ func (s *Server) heartbeat(ctx context.Context) <-chan error {
 	s.tg.Go("heartbeat", func() {
 		defer func() {
 			close(errc)
-			s.Close()
+			defer func() { go s.Close() }()
 		}()
 
 		for {
@@ -251,7 +251,7 @@ func (s *Server) Start() error {
 	}()
 
 	s.tg.GoNoPanic("collect", func() {
-		defer s.Close()
+		defer func() { go s.Close() }()
 		s.collector.Start(s.ctx)
 	})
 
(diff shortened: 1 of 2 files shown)


suzaku:
Add a integration test case for `recover table`?

july2993:
> Add a integration test case for `recover table`?

for it to successfully exec at downstream, we must change the downstream tidb to use tikv instead, i will ask @lichunzhu for help to add this  integration test after this pr, is this ok?

july2993:
@WangXiangUSTC PTAL

suzaku:
LGTM

drainer/server.go
```diff
@@ -216,7 +216,7 @@ func (s *Server) heartbeat(ctx context.Context) <-chan error {
 	s.tg.Go("heartbeat", func() {
 		defer func() {
 			close(errc)
-			s.Close()
+			defer func() { go s.Close() }()
```
suzaku: Can be just `go s.Close()`?

july2993: update by https://github.com/pingcap/tidb-binlog/pull/674/commits/ad0a4a08a308d0e1ee6e994d82cdc4a0249b1396

drainer/server.go
```diff
@@ -251,7 +251,7 @@ func (s *Server) Start() error {
 	}()
 
 	s.tg.GoNoPanic("collect", func() {
-		defer s.Close()
+		defer func() { go s.Close() }()
```
WangXiangUSTC: it is ok now, but I think maybe it is better to define a new function like `Stop` to do some thing in `Close`, and only the main thread call the `Close` function to do `s.tg.Wait()`

Issue #665, opened 2019-07-03
drainer create table error if sql include"\n"


jwongz:
version:v2.1.4
syncer.to  mysql(is a TiDB)

sql:
```
CREATE TABLE films (
  id int(11),
  release_year int(11),
  category_id int(11),
  rating decimal(3,2)
);
insert into films values
(1,2015,1,8.00),
(2,2015,2,8.50),
(3,2015,3,9.00),
(4,2016,2,8.20),
(5,2016,1,8.40),
(6,2017,2,7.00);
```

error log :
```
{"log":"2019/07/03 16:44:08 pump.go:115: \u001b[0;37m[info] [pump 13274bc26df3:10060] create pull binlogs client\u001b[0m\n","stream":"stderr","time":"2019-07-03T08:44:08.719759354Z"}
{"log":"2019/07/03 16:44:08 pump.go:115: \u001b[0;37m[info] [pump 33576518e473:10060] create pull binlogs client\u001b[0m\n","stream":"stderr","time":"2019-07-03T08:44:08.719763484Z"}
{"log":"2019/07/03 16:44:08 pump.go:115: \u001b[0;37m[info] [pump 7410ddec094a:10060] create pull binlogs client\u001b[0m\n","stream":"stderr","time":"2019-07-03T08:44:08.719767392Z"}
{"log":"2019/07/03 16:44:09 syncer.go:503: \u001b[0;37m[info] [ddl][start]use `binlog`; CREATE TABLE films (\n","stream":"stderr","time":"2019-07-03T08:44:09.669498645Z"}
{"log":"  id int(11),\n","stream":"stderr","time":"2019-07-03T08:44:09.669534124Z"}
{"log":"  release_year int(11),\n","stream":"stderr","time":"2019-07-03T08:44:09.669542576Z"}
{"log":"  category_id int(11),\n","stream":"stderr","time":"2019-07-03T08:44:09.669549275Z"}
{"log":"  rating decimal(3,2)\n","stream":"stderr","time":"2019-07-03T08:44:09.669581963Z"}
{"log":");[commit ts]409506059518476290\u001b[0m\n","stream":"stderr","time":"2019-07-03T08:44:09.669588491Z"}
{"log":"2019/07/03 16:44:09 sql.go:107: \u001b[0;31m[error] exec sqls[[use `binlog`; CREATE TABLE films (\n","stream":"stderr","time":"2019-07-03T08:44:09.679585095Z"}
[...]

PR #677, opened 2019-07-12
Split composite DDL like "use <db>; <ddl>;" to improve DDL compatibility

lichunzhu:
[#665](https://github.com/pingcap/tidb-binlog/issues/665)
[TOOL-1375](https://internal.pingcap.net/jira/browse/TOOL-1375)
Some sql proxies don’t support SQL like `use <db>; <ddl>`.

* This PR will add args for DDLs of database name and table name. For function `GenDDL()` now the DDL sql will be returned directly.
* Before execution, we will check whether this sql is a DDL or not. If it is and has index for table name, we will add a `use <db>;` command before the actual sql. For all DDLs their args will be clean before execution.
* The `[]sqls` of release-2.1 are executed by a "transactional" way, which means we use `tx.begin()` and `tx.commit()` to execute the DDL. But the tranction of DDL can't be guaranteed. The main benifit is that in this way can hold only one `*sql.conn` and the execution order can be assured so that we can split the sql into two parts.

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

Side effects

Related changes


drainer/translator/mysql.go
```diff
@@ -664,3 +664,18 @@ func DecodeOldAndNewRow(b []byte, cols map[int64]*types.FieldType, loc *time.Loc
...
 	return oldRow, newRow, nil
 }
+
+// SplitWithSemicolons is used to make it compatible with some SQL
+// that cannot execute multiple statements at once
+func SplitWithSemicolons(sqls []string, args [][]interface{}) ([]string, [][]interface{}) {
+	nsqls := make([]string, 0)
+	nargs := make([][]interface{}, 0)
+	for i, sql := range sqls {
+		s := strings.SplitAfter(sql, ";")
```
kennytm: What happens when the SQL itself contains a `;`... e.g.

```sql
CREATE TABLE `foo;bar` (`col;umn` VARCHAR(20) DEFAULT 'va;lue') COMMENT = 'com;ment';
```

lichunzhu: Actually the program will fail. Maybe we can just check these semicolons just before some key words? Such as `USE, SELECT, CREATE, INSERT`...etc.

lichunzhu: Or we can check the semicolon(;) in quotation marks, like `a;b`. Regard these kinds of semicolon as a part of name.

lichunzhu: Or we can check the semicolon(;) in quotation marks, like \`a;b\`. Regard these kinds of semicolon as a part of name.

GregoryIan:
what's meaning of `The []sqls of release-2.1 are executed by transaction`, same with `DDL` statement?

drainer/translator/mysql.go
GregoryIan: it's not so simple

GregoryIan:
and please add labels for PR

lichunzhu:
@GregoryIan The description is not precise. The DDL will be executed in a 'transactional' way, which means we use `tx.begin()` and `tx.commit()` to execute the DDL. But the tranction of DDL can't be guaranteed. 
The main benifit is that this way can hold only one `*sql.conn` so that we can split the sql into two parts, in my opinion.

GregoryIan:
I understand your description now. do you know what's the problem of this pr now? and how do you want to solve it? @lichunzhu

lichunzhu:
@GregoryIan I understand why we can't just simply split the DDL with semicolons. Now I'm trying to read `/reparo/syncer/translate.go:parserSchemaTableFromDDL()` to figure out a way to solve this problem with the help of parser.

drainer/executor/mysql.go
```diff
@@ -30,6 +30,15 @@ func newMysql(cfg *DBConfig, sqlMode *string) (Executor, error) {
 }
 
 func (m *mysqlExecutor) Execute(sqls []string, args [][]interface{}, commitTSs []int64, isDDL bool) error {
+	if isDDL {
+		// the index of table is not empty
+		if args[0][1] != "" {
```
july2993: why check table but not schema?

drainer/executor/mysql.go
```diff
@@ -30,6 +30,15 @@ func newMysql(cfg *DBConfig, sqlMode *string) (Executor, error) {
...
+			schema := args[0][0]
+			useSQL := fmt.Sprintf("use %s;", schema)
```
july2993: add `use <db>` only if it's not a create table statement like in `GenDDLSQL` before.
or will fail to exec `use <db>` because the downstream has not the <db>

drainer/translator/mysql.go
```diff
@@ -275,19 +273,7 @@ func (m *mysqlTranslator) genDeleteSQL(schema string, table *model.TableInfo, co
...
-		return "", errors.Trace(err)
-	}
-
-	_, isCreateDatabase := stmt.(*ast.CreateDatabaseStmt)
-	if isCreateDatabase {
-		return fmt.Sprintf("%s;", sql), nil
-	}
-
-	return fmt.Sprintf("use `%s`; %s;", schema, sql), nil
+	return fmt.Sprintf("%s;", sql), nil
```
july2993: just return sql, nil
if you want <sql>;  you can just use `sql + ";" here for more clear and performance

drainer/executor/mysql.go
lichunzhu: revealed

drainer/translator/mysql.go
lichunzhu: revealed

drainer/executor/mysql.go
```diff
@@ -13,23 +16,49 @@ import (
...
+			return errors.Trace(err)
+		}
+
+		_, isCreateDatabase := stmt.(*ast.CreateDatabaseStmt)
+		if !isCreateDatabase {
+			schema := args[0][0]
+			useSQL := fmt.Sprintf("use %s;", schema)
+			sqls = append([]string{useSQL}, sqls...)
+		}
+		args = make([][]interface{}, len(sqls))
```
suzaku: Is it safe to override `args` here?

drainer/syncer.go
```diff
@@ -517,7 +517,7 @@ func (s *Syncer) run(jobs []*model.Job) error {
 				var args []interface{}
 				// for kafka, we want to know the relate schema and table, get it while args now
```
suzaku: Should also update the comment.

lichunzhu:
Two more comments were added. PTAL. @suzaku

PR #678, opened 2019-07-15
tests: No need to wait for seconds before closing

suzaku:

Make `TestCloseAndOpenAgain` -- which is one of the slowest test case --
faster.

We had to wait for about 3 seconds before closing because
`handleSortItems` only save maxCommitTS and handlePointer at most
once per second. But when testing, we don't want to wait for so long.

Make the interval an variable, so that we can override it with shorter
durations in the test case.

Now we are waiting milliseconds instead of seconds.

### Check List 

Tests 

 - Unit test

Code changes

Side effects

Related changes


leoppro:
LGTM

PR #685, opened 2019-07-22
drainer/: Add ignore-txn-commit-ts to skip some txn

july2993:
add ignore-txn-commit-ts to skip some txn.
use case: for some incompatible DDL, user may handle it manaually and want to skip this binlog, currentlly, usert can only change the checkpoint to skip this,  the process is tedious and prone to error.

add `ignore-txn-commit-ts` support to skip specified commit ts txn

### Check List 

Tests 

 - Unit test
 - Integration test
config `ignore-txn-commit-ts` and make sure it skip it

Code changes

Side effects

Related changes

### DIFF
--- a/drainer/syncer.go
+++ b/drainer/syncer.go
@@ -400,7 +400,7 @@ ForLoop:
 				s.itemsWg.Add(1)
 				lastAddComitTS = binlog.GetCommitTs()
 
-				log.Info("add ddl item to syncer, you can add this commit ts to `ignore-txn-commit-ts` for skip this ddl if need",
+				log.Info("add ddl item to syncer, you can add this commit ts to `ignore-txn-commit-ts` to skip this ddl if needed",
 					zap.String("sql", sql), zap.Int64("commit ts", binlog.CommitTs))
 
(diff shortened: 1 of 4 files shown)


drainer/syncer.go
```diff
@@ -394,6 +399,10 @@ ForLoop:
 				beginTime := time.Now()
 				s.itemsWg.Add(1)
 				lastAddComitTS = binlog.GetCommitTs()
+
+				log.Info("add ddl item to syncer, you can add this commit ts to `ignore-txn-commit-ts` for skip this ddl if need",
```
suzaku: Maybe only output this line when `Sync` failed in `if err != nil`?

july2993: it's ok to log info for every ddl binlog statement(this is important and will not too much).
but it better only log the part `you can add this commit ts to ignore-txn-commit-ts..` only when failing to exec the ddl, but the `Sync` is async, it returns success only means add to syncer success `but not` successfully synced to downstream, so log it for every ddl now.

kennytm:
```suggestion
				log.Info("add ddl item to syncer, you can add this commit ts to `ignore-txn-commit-ts` to skip this ddl if needed",
```

PR #687, opened 2019-07-23
Return error when loading some undefined configs from local

lichunzhu:
[TOOL-1403](https://internal.pingcap.net/jira/browse/TOOL-1403)
Referral PR: [TiDB#9855](https://github.com/pingcap/tidb/pull/9855/)
Now func `configFromFile` in `config.go` of drainer, pump, arbiter, reparo will make sure that there is no undecoded args in `metaData`. This can help customers check their misspell in the config toml files. Relative tests have been added, too.

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

Side effects

Related changes


drainer/config_test.go
```diff
@@ -118,3 +121,39 @@ func (t *testDrainerSuite) TestAdjustConfig(c *C) {
...
+		true,
+	}
+
+	var buf bytes.Buffer
+	e := toml.NewEncoder(&buf)
+	err := e.Encode(yc)
+	c.Assert(err, IsNil)
+
+	tmpfile, err := util.CreateCfgFile(buf.Bytes(), "drainer_config")
+	if tmpfile != nil && len(tmpfile.Name()) > 0 {
```
lichunzhu: I'm not so sure about this part. But I think it's necessary to delete tmpfile once this file was created, even if some error occured.

drainer/config.go
```diff
@@ -223,8 +223,21 @@ func (c *SyncerConfig) adjustDoDBAndTable() {
...
+
+	// If any items in confFile file are not mapped into the Config struct, issue
+	// an error and stop the server from starting.
+	if err == nil {
+		if undecoded := metaData.Undecoded(); len(undecoded) > 0 {
+			var undecodedItems []string
+			for _, item := range undecoded {
+				undecodedItems = append(undecodedItems, item.String())
+			}
+			err = errors.New(fmt.Sprintf("config file %s contained unknown configuration options: %s", path, strings.Join(undecodedItems, ", ")))
```
suzaku: You can use `errors.Errorf` here.

drainer/config_test.go
suzaku: You can use `c.Mkdir()` to create a temporary directory and let gocheck do it for you.

drainer/config_test.go
```diff
@@ -118,3 +121,39 @@ func (t *testDrainerSuite) TestAdjustConfig(c *C) {
...
+		UnrecognizedOptionTest bool   `toml:"unrecognized-option-test" json:"unrecognized-option-test"`
+	}{
+		"data.drainer",
+		"192.168.15.10:8257",
+		"192.168.15.10:8257",
+		true,
+	}
+
+	var buf bytes.Buffer
+	e := toml.NewEncoder(&buf)
```
suzaku: You may use a string instead, for example: https://github.com/pingcap/dm/pull/212/files#diff-fa28104e876d9e8e832e695718aa8edeR23

drainer/config.go
```diff
@@ -223,8 +223,21 @@ func (c *SyncerConfig) adjustDoDBAndTable() {
 }
 
 func (cfg *Config) configFromFile(path string) error {
-	_, err := toml.DecodeFile(path, cfg)
-	return errors.Trace(err)
+	metaData, err := toml.DecodeFile(path, cfg)
+
+	// If any items in confFile file are not mapped into the Config struct, issue
+	// an error and stop the server from starting.
+	if err == nil {
```
july2993: handle err first, like:
```
if err != nil {
    return ...
}
// handle normal case
```

pkg/util/util.go
```diff
@@ -258,3 +260,23 @@ func AdjustDuration(v *time.Duration, defValue time.Duration) {
 		*v = defValue
 	}
 }
+
+// CreateCfgFile creates and writes b as config file to dir prefix
+func CreateCfgFile(b []byte, prefix string) (*os.File, error) {
```
july2993: https://golang.org/pkg/io/ioutil/#WriteFile

drainer/config_test.go
lichunzhu: Revised yet.

drainer/config.go
lichunzhu: Revised.

drainer/config_test.go
lichunzhu: I think struct is better for comparison when the demand changes so I did't change the code here.

pkg/util/util.go
```diff
@@ -258,3 +260,23 @@ func AdjustDuration(v *time.Duration, defValue time.Duration) {
 		*v = defValue
 	}
 }
+
+// CreateCfgFile creates and writes b as config file to dir prefix
+func CreateCfgFile(b []byte, dir, prefix string) (*os.File, error) {
```
suzaku: This is an example of how a file can be written for testing: https://github.com/pingcap/dm/pull/212/files#diff-fa28104e876d9e8e832e695718aa8edeR86-R98

See if you can rewrite the test so that you don't need to add this new function.

pump/config_test.go
```diff
@@ -116,21 +117,46 @@ func (s *testConfigSuite) TestConfigParsingFileFlags(c *C) {
...
-	mustSuccess(c, err)
+	args := []string{
+		"--config",
+		configFilename,
+		"-L", "debug",
+	}
 
-	err = tmpfile.Close()
-	mustSuccess(c, err)
+	os.Clearenv()
```
suzaku: Why do we need to clear environment variables before parsing again?

lichunzhu: It will be deleted in next commit.

pkg/util/util.go
lichunzhu: Revised.

pkg/util/util.go
lichunzhu: Revised.

drainer/config.go
lichunzhu: Revised.

pump/config_test.go
lichunzhu: In `pump/config_test.go:75` test TestConfigParsingEnvFlags set environment variables. Now I add a defer clean job in this function. The other functions' os.Clearenv() have already been expurgated.

arbiter/config.go
```diff
@@ -190,6 +191,20 @@ func (cfg *Config) adjustConfig() error {
...
+	// an error and stop the server from starting.
+	if err != nil {
+		return err
+	}
+	if undecoded := metaData.Undecoded(); len(undecoded) > 0 {
+		var undecodedItems []string
+		for _, item := range undecoded {
+			undecodedItems = append(undecodedItems, item.String())
+		}
+		err = errors.Errorf("arbiter config file %s contained unknown configuration options: %s", path, strings.Join(undecodedItems, ", "))
```
suzaku: Maybe we can create a function in `pkg/util` like `StrictDecodeFile` and then reuse it in the `configFromFile` methods of different packages?

lichunzhu: Nice suggestion. I will take a try.

suzaku: Great :)

lichunzhu: Updated.

reparo/config.go
```diff
@@ -17,11 +17,11 @@ import (
 	"encoding/json"
 	"flag"
 	"fmt"
+	"github.com/pingcap/tidb-binlog/pkg/util"
```
kennytm: This import should be placed in the next section

reparo/config_test.go
```diff
@@ -14,7 +14,10 @@
 package reparo
 
 import (
+	"bytes"
 	"fmt"
+	"github.com/BurntSushi/toml"
```
kennytm: ditto

arbiter/config_test.go
```diff
@@ -14,7 +14,10 @@
 package arbiter
 
 import (
+	"bytes"
 	"fmt"
+	"github.com/BurntSushi/toml"
```
kennytm: ditto

drainer/config_test.go
```diff
@@ -14,6 +14,10 @@
 package drainer
 
 import (
+	"bytes"
+	"github.com/BurntSushi/toml"
```
kennytm: ditto

pkg/util/util.go
```diff
@@ -16,7 +16,9 @@ package util
 import (
 	"context"
 	"fmt"
+	"github.com/BurntSushi/toml"
```
kennytm: ditto

pump/config_test.go
```diff
@@ -15,11 +15,11 @@ package pump
...
 import (
 	"bytes"
-	"io/ioutil"
-	"os"
-
 	"github.com/BurntSushi/toml"
 	. "github.com/pingcap/check"
+	"io/ioutil"
+	"os"
+	"path"
```
kennytm: These 3 imports should be placed in the previous section

arbiter/config.go
```diff
@@ -17,14 +17,13 @@ import (
...
-	"os"
-
-	"github.com/BurntSushi/toml"
 	"github.com/pingcap/errors"
 	"github.com/pingcap/log"
 	"github.com/pingcap/tidb-binlog/pkg/flags"
+	"github.com/pingcap/tidb-binlog/pkg/util"
 	"github.com/pingcap/tidb-binlog/pkg/version"
 	"go.uber.org/zap"
+	"os"
```
kennytm: ditto

PR #691, opened 2019-07-25
loader: stop retry execute sqls when closed 

WangXiangUSTC:
fix issue: https://internal.pingcap.net/jira/browse/TOOL-1018
when execute sqls meet error and user kill drainer, drainer will not exit immediately and will still retry util reach the max retry time.

1. add stopRetry function as argument in RetryOnError function
2. when loader is closed, stop retry

### Check List 

Tests 

 - Unit test
 - Manual test (when drainer execute sql and meet error, kill drainer, check drainer is exit or not)

Related changes


kennytm:
Why not use `RetryContext` (and cancel a Context on exit)?

WangXiangUSTC:
do you mean add a context in `loader`, and judge ctx.Done() when retry? But it is not suggest to store contexts inside a struct.  @kennytm

kennytm:
Besides storing the context into the struct, execDDL could accept a context as argument.

WangXiangUSTC:
it is difficult to use context as argument in this situation now @kennytm

WangXiangUSTC:
if so we need add a context in loaderImpl, do you think it's ok? @suzaku

suzaku:
> if so we need add a context in loaderImpl, do you think it's ok? @suzaku

I think it's OK in this case because it could enable the reuse of `RetryContext` so that we don't need to change `RetryOnError`.

WangXiangUSTC:
code updated, but I think we may need refine it someday, it is not suggest to store context in struct. @kennytm @suzaku PTAL again

kennytm:
Hmm I don't know, what I had in mind is make it `execDDL(Context, *DDL)` up to `(*loaderImpl) Run(Context)`, and when the program stops the Context is canceled externally.

WangXiangUSTC:
drainer/server -> drainer/syncer -> drainer/sync/MysqlSyncer -> pkg/Loader, so we need add context in MysqSyncer? or drainer/server?  and loader already provide a Close() function, why need do cancel context again? @kennytm

kennytm:
@WangXiangUSTC basically add Context to every function 🤪. So you don't need to store Context in the struct. Implementation-wise LGTM but I'm not sure about the design. WDYT @suzaku

suzaku:
@kennytm I think these models as a whole were not designed to work well with context, maybe we can improve this with another PR later. For now, allowing `loaderImpl` to manage its own lifetime `context` should be the solution that requires minimal change.

suzaku:
I just tried to refactor these functions to make them `context`-aware, but then I noticed that I also had to refactor the caller of `Loader`, it would be a lot of work.
A bigger problem is that we didn't design the `Close` methods very well in the first place, so in many cases, many cooperation need to happen for both the caller and the callee to quit correctly.

BTW, @WangXiangUSTC , there are many `Exec` calls under the hood, shall we also replace them with `ExecContext`?

WangXiangUSTC:
@suzaku updated in https://github.com/pingcap/tidb-binlog/pull/691/commits/694765280940c2825c0043c2f31071e5d05a8f64 and https://github.com/pingcap/tidb-binlog/pull/691/commits/076a4558c46cd1c56b32be3867d6b970e83b859d

pkg/loader/load.go
```diff
@@ -86,6 +86,9 @@ type loaderImpl struct {
 	// value can be tidb or mysql
 	saveAppliedTS           bool
 	lastUpdateAppliedTSTime time.Time
+
+	ctx    context.Context
```
kennytm: Add a TODO that we shouldn't store the `ctx` in the struct and should be refactored in the future etc.

pkg/loader/load.go
```diff
@@ -222,6 +232,7 @@ func (s *loaderImpl) Successes() <-chan *Txn {
 // Run will quit when all data is drained
 func (s *loaderImpl) Close() {
```
july2993:
```
221 // Close close the Loader, no more Txn can be push into Input()
222 // Run will quit when all data is drained
223 func (s *loaderImpl) Close() {
224     close(s.input)
225 }
```
what's the behavior now?

pkg/loader/load.go
```diff
@@ -222,6 +232,7 @@ func (s *loaderImpl) Successes() <-chan *Txn {
...
 	close(s.input)
+	s.cancel()
```
july2993: in `Run()` it will still execute all the accumulated dmls (but with the context that has been canceled?)

WangXiangUSTC: already remove ctx when execute sqls

WangXiangUSTC:
@july2993 @kennytm PTAL again

pkg/loader/load_test.go
```diff
@@ -58,6 +59,7 @@ func (cs *LoadSuite) TestGetExecutor(c *check.C) {
 		metrics: &MetricsGroup{
 			QueryHistogramVec: &prometheus.HistogramVec{},
 		},
+		ctx: context.Background(),
```
kennytm: The `cancel` field needs to be non-null to avoid `s.cancel()` crashing, right?

WangXiangUSTC: yes

WangXiangUSTC: but the unit test dosen't call loader.Close(), no crashing now

july2993:
@suzaku PTAL and approve it if still LGTM for you

PR #695, opened 2019-07-26
binlogctl: Use registry.Node to get a single node directly

suzaku:

`registry.Nodes` is used to get all nodes before searching for the node
with the specified ID.

Use `registry.Node` to look up the specified node ID directly.

The error output for nodes not found is a little different from before,
since the error is logged directly.

### Check List 

Tests 

 - Unit test
 - Manual test

Code changes

Side effects

Related changes

### DIFF
--- a/binlogctl/nodes_test.go
+++ b/binlogctl/nodes_test.go
@@ -23,10 +23,9 @@ import (
 	"testing"
 	"time"
 
-	"github.com/pingcap/errors"
-
 	"github.com/coreos/etcd/integration"
 	. "github.com/pingcap/check"
+	"github.com/pingcap/errors"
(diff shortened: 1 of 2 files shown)


binlogctl/nodes_test.go
```diff
@@ -23,6 +23,8 @@ import (
 	"testing"
 	"time"
 
+	"github.com/pingcap/errors"
+
```
WangXiangUSTC: can remove this line

july2993:
@suzaku pls add labels

binlogctl/nodes_test.go
suzaku: Fixed.

PR #697, opened 2019-08-01
drainer/: Add ignore-txn-commit-ts to skip some txn (#685)

july2993:
add ignore-txn-commit-ts to skip some txn.
use case: for some incompatible DDL, user may handle it manaually and want to skip this binlog, currentlly, usert can only change the checkpoint to skip this, the process is tedious and prone to error.
cherry-pick https://github.com/pingcap/tidb-binlog/pull/685

### Check List 

Tests 

 - Unit test

Code changes

Side effects

Related changes


lichunzhu:
LGTM

PR #698, opened 2019-08-05
Makefile: Version get the latest v4.0 tag for master

july2993:
Version get the latest v4.0 tag for master
so we can keep release3.0 and master don't diverse and show different version for `-V`
add `-match 'v4.0*` opt for git description

### Check List 

Tests 

 - No code

Code changes

Side effects

 - Related changes


lichunzhu:
LGTM

kennytm:
>  so we can keep release3.0 and master don't diverse and show different version for -V

How would this work on the release-3.0 branch in the future?

july2993:
> > so we can keep release3.0 and master don't diverse and show different version for -V
> 
> How would this work on the release-3.0 branch in the future?

we can change release-3.0 to match the v3.0 too if need

kennytm:
So release-3.0 and master should diverge after this commit?

july2993:
> So release-3.0 and master should diverge after this commit?

💔

july2993:
looks we need to check the current branch in a `if .. else ..` way

GregoryIan:
@leoppro PTAL

PR #701, opened 2019-08-05
pump/: Fix block at wait for MaxCommitTS when offline

july2993:
when offline:
1, ApplyAction change statue to be closing, reject write but accept fake
binlog
2, commitStatus() -> waitSafeToOffline(), we will write a fake binlog
and wait when storage.MaxCommitTS() >= commit ts of the fake binlog
...

after 1, may having some unwritten c-binlog(we will reject the write
request from tidb), and in sorter, block at s.cond.Wait case because no
more any item push.

only reject p-binlog when not online, and signal to wait up in sorter period.

### Check List 

Tests 

 - Unit test
 - Integration test
offline pump when write qps = 10k

Code changes

Side effects

 Related changes

### DIFF
--- a/pump/server.go
+++ b/pump/server.go
@@ -788,7 +788,7 @@ func (s *Server) waitSafeToOffline(ctx context.Context) error {
 
 	log.Debug("Start waiting until all drainers have consumed the last fake binlog")
 
-	waterMark := s.storage.MaxCommitTS()
+	maxCommitTS := s.storage.MaxCommitTS()
 	for {
 		select {
 		case <-time.After(time.Second):
@@ -797,12 +797,12 @@ func (s *Server) waitSafeToOffline(ctx context.Context) error {
 				log.Error("Failed to get safe GCTS", zap.Error(err))
 				break
 			}
-			if safeTSO >= waterMark {
+			if safeTSO >= maxCommitTS {
 				return nil
 			}
 			log.Warn("Waiting for drainer to consume binlog",
 				zap.Int64("Minimum Drainer MaxCommitTS", safeTSO),
-				zap.Int64("Water mark CommiTS", waterMark))
+				zap.Int64("Need to reach maxCommitTS", maxCommitTS))
 			if _, err = s.writeFakeBinlog(); err != nil {
 				log.Error("write fake binlog failed", zap.Error(err))
 			}
--- a/pump/server.go
+++ b/pump/server.go
@@ -216,7 +216,7 @@ func (s *Server) writeBinlog(ctx context.Context, in *binlog.WriteBinlogReq, isF
 		goto errHandle
 	}
 
-	if !isFakeBinlog {
+	if !isFakeBinlog && blog.Tp == binlog.BinlogType_Prewrite {
 		state := s.node.NodeStatus().State
 		if state != node.Online {
 			err = errors.Errorf("no online: %v", state)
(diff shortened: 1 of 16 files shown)


pump/server.go
```diff
@@ -216,7 +216,7 @@ func (s *Server) writeBinlog(ctx context.Context, in *binlog.WriteBinlogReq, isF
 		goto errHandle
 	}
 
-	if !isFakeBinlog {
+	if !isFakeBinlog && blog.Tp == binlog.BinlogType_Prewrite {
```
suzaku: In the original implementation, when closing, a fake binlog is written and assumed to be the last binlog. This is true, because in this line, we guaranteed that no other non-fake binlogs could be written.

With this change, this assumption no longer holds, for example, in the following case, the commit binlog may get lost:

1. Close
1. Write Fake binlog
1. A C-Binlog is written
1. All online drainers have consumed the fake binlog but not the C-Binlog
1. The Pump is allowed to go offline and the drainers would not observe this missing C-Binlog when merging.

july2993: add a method to make sure all p-binlog having matched c-binlog PTAL

pump/storage/sorter.go
```diff
@@ -149,6 +159,18 @@ func (s *sorter) pushTSItem(item sortItem) {
 func (s *sorter) run() {
 	defer s.wg.Done()
 
+	go func() {
+		// Avoid if no any more pushTSItem call so block at s.cond.Wait() in run() waiting the matching c-binlog
```
suzaku: When `sorter` get closed, `s.cond.Broadcast` is called to wake all goroutines waiting for `s.cond`, isn't that enough?

pump/server.go
```diff
@@ -788,6 +788,7 @@ func (s *Server) waitSafeToOffline(ctx context.Context) error {
 
 	log.Debug("Start waiting until all drainers have consumed the last fake binlog")
 
+	waterMark := s.storage.MaxCommitTS()
```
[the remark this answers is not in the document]
july2993: 水位标记

pump/storage/sorter.go
july2993: we don't close it, it may block at s.cond.Wait case when no
more any item push.

suzaku: When closing the server, storage is closed and so is the sorter in it.

july2993: not talking the closing the server case, if block at the no matching c-binlog case it will nerver wake up if not any more any item push (tidb may crash and no c-binlog write, we may stop to write fakebinlog when offline in waitUntilCommitTSSaved)

suzaku: I see.

But with the following goroutine added, we will be periodically waking up `Wait`s, which is just like when we were checking periodically. Then maybe we should just get rid of `s.cond` because it's no longer helping us?

july2993: it can still make it wake up immediately once we push new item anc check for matching c-binog but not just check periodically

pump/storage/sorter.go
```diff
@@ -124,6 +124,16 @@ func (s *sorter) setResolver(resolver func(startTS int64) bool) {
...
 }
 
+func (s *sorter) allMatched() bool {
+	var match bool
+
+	s.lock.Lock()
+	match = len(s.waitStartTS) == 0
+	s.lock.Unlock()
+
+	return match
```
suzaku: How about using `defer` and return the bool expression directly?

july2993: https://github.com/pingcap/tidb-binlog/pull/701/commits/9c56122d3a6ad56ab107022265301ac87ea28b73

PR #702, opened 2019-08-07
Add metric for batch write kv

leoppro:
In some extreme cases, when leveldb write failed, pump server may be entering an infinite loop

### Check List 

Tests 

 - No Test


leoppro:
@july2993 @suzaku PTAL

PR #712, opened 2019-08-13
*: Fix if only one column and pk is not handle

july2993:
columnValues will be nil and panic

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes


GregoryIan:
lgtm

PR #714, opened 2019-08-14
fix bug when some error occurs in loader arbiter's syncBinlog may block

lichunzhu:
[TOOL-1488](https://internal.pingcap.net/jira/browse/TOOL-1488)
When loader executes some dmls and meet an error in `loader.Run()`, it will shut down, report an error and stop dealing with chan `loader.input`. But `syncBinlogs` in `arbiter/server.go` will keep on pushing items into `loader.input`, when it get full, function `syncBinlogs` will be blocked and then arbiter will be unable to quit or issue the error.
Add `syncCtx` and `select` in function `syncBinlogs` to avoid blocking.

### Check List 

Tests 

 - Unit test
 - Integration test
Modified arbiter will quit as we expect.
image

Code changes

Side effects

Related changes

### DIFF
--- a/arbiter/server.go
+++ b/arbiter/server.go
@@ -162,13 +162,7 @@ func (s *Server) Close() error {
 
 // Run runs the Server, will quit once encounter error or Server is closed
 func (s *Server) Run() error {
-	defer func() {
-		if s.downDB != nil {
-			s.downDB.Close()
-		} else {
-			log.Error("Invalid downDB address, please check the config")
-		}
-	}()
+	defer s.downDB.Close()
 
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
--- a/arbiter/server.go
+++ b/arbiter/server.go
@@ -162,7 +162,13 @@ func (s *Server) Close() error {
 
 // Run runs the Server, will quit once encounter error or Server is closed
 func (s *Server) Run() error {
-	defer s.downDB.Close()
+	defer func() {
+		if s.downDB != nil {
+			s.downDB.Close()
+		} else {
+			log.Error("Invalid downDB address, please check the config")
+		}
+	}()
 
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
--- a/arbiter/server.go
+++ b/arbiter/server.go
@@ -200,7 +200,6 @@ func (s *Server) Run() error {
 
 	wg.Wait()
 
-	// to pass go check
 	syncCancel()
 	if err != nil {
 		return errors.Trace(err)
--- a/arbiter/server.go
+++ b/arbiter/server.go
@@ -186,7 +186,7 @@ func (s *Server) Run() error {
 	wg.Add(1)
 	go func() {
 		defer wg.Done()
-		syncErr = syncBinlogs(s.kafkaReader.Messages(), s.load, syncCtx)
+		syncErr = syncBinlogs(syncCtx, s.kafkaReader.Messages(), s.load)
 		if syncErr != nil {
 			s.Close()
 		}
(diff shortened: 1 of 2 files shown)


arbiter/server_test.go
```diff
@@ -407,7 +407,7 @@ func (s *syncBinlogsSuite) TestShouldSendBinlogToLoader(c *C) {
 	}()
 	ld := dummyLoader{input: dest}
 
-	err := syncBinlogs(source, &ld)
+	err := syncBinlogs(context.Background(), source, &ld)
```
suzaku: Add a test case here to show that `syncBinlogs` can be canceled from outside?

lichunzhu: I add another test case in the end

arbiter/server.go
```diff
@@ -182,22 +188,26 @@ func (s *Server) Run() error {
...
+		syncErr = syncBinlogs(syncCtx, s.kafkaReader.Messages(), s.load)
 		if syncErr != nil {
 			s.Close()
 		}
 	}()
 
 	err := s.load.Run()
 	if err != nil {
+		syncCancel()
 		s.Close()
```
suzaku: Maybe we can fix this problem in a simpler way? For example, it seems buggy that `Reader.Close` doesn't `close` the `Messages` channel.

lichunzhu: `Reader.Close` has closed successfully, but `Reader.msgs` channel has a buffer size of 1024. Func `syncCancel` can still read the buffered `msg` although it is closed.

arbiter/server_test.go
```diff
@@ -418,3 +418,38 @@ func (s *syncBinlogsSuite) TestShouldSendBinlogToLoader(c *C) {
...
+func (s *syncBinlogsSuite) TestShouldQuitWhenSomeErrorOccurs(c *C) {
+	readerMsgs := make(chan *reader.Message, 1024)
+	dummyLoaderImpl := &dummyLoader{
+		successes: make(chan *loader.Txn),
+		// input is set small to trigger blocking easily
+		input: make(chan *loader.Txn, 1),
+	}
+	msg := s.createMsg("test42", "users", "alter table users add column gender smallint")
+	ctx, cancel := context.WithCancel(context.Background())
+	syncCtx, syncCancel := context.WithCancel(ctx)
```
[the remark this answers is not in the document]
lichunzhu: Revised

lichunzhu:

arbiter/server.go
```diff
@@ -162,7 +162,13 @@ func (s *Server) Close() error {
 
 // Run runs the Server, will quit once encounter error or Server is closed
 func (s *Server) Run() error {
-	defer s.downDB.Close()
+	defer func() {
+		if s.downDB != nil {
+			s.downDB.Close()
+		} else {
+			log.Error("Invalid downDB address, please check the config")
```
suzaku: When would `downDB` be `nil`?

lichunzhu: Revised. This part was modified because at first I wanna test the whole `Run` function.

arbiter/server.go
```diff
@@ -182,22 +182,26 @@ func (s *Server) Run() error {
...
 
 	err := s.load.Run()
 	if err != nil {
+		syncCancel()
 		s.Close()
 	}
 
 	wg.Wait()
 
+	// to pass go check
```
july2993: address the comment, we don't write code for static check tools... the problem is why we need to cancel here?
just remove the comment is better to add a "to pass go check"

lichunzhu:
@july2993 There is a bug in the previous test. In `readerMsgs <- msg` when `ctx` is cancelled directly there might be no `msg` in channel `readerMsgs` and `syncBinlogs` will get blocked since channel `readerMsgs` is not closed. Now 3 msgs will be sent to `readerMsgs` and after `ctx` is cancelled `readerMsgs` will be closed. PTAL.

PR #709, opened 2019-08-12
Pump: Fix pump server start process

leoppro:
### What problem does this PR solve? 
pump server start process:
1. register the node in pd
....
n. listen ports(http and grpc service)

This leads to pump client maybe find an inaccessible pump server in pd(not listen ports yet)

### What is changed and how it works?

change the pump server start process: first, listen ports and then register the node

### Check List 

Tests 

 - Unit test
 - Integration test
 - Manual test (add detailed scripts or steps below)
 - No code

Code changes

 - Has exported function/method change
 - Has exported variable/fields change
 - Has interface methods change
 - Has persistent data change

Side effects

 - Possible performance regression
 - Increased code complexity
 - Breaking backward compatibility

Related changes

 - Need to cherry-pick to the release branch
 - Need to update the documentation
 - Need to update the `tidb-ansible` repository
 - Need to be included in the release note

PR #716, opened 2019-08-15
fix pump init order, register the node after listen ports (#709)

leoppro:
#### What problem does this PR solve? 

cherry-pick #709 to release-2.1


leoppro:
@suzaku PTAL

PR #718, opened 2019-08-16
Avoid repeated called of NewIterator which might be slow

leoppro:
When the size of metadata(storage with leveldb) grown very large, pump storage gc will be very slowly. The reason for this bug is the `seek` operation in `goleveldb` spend too much time.

we `seek` only one time on the `init` process of GC, instead of `seek` every time before `vlog gc`.
### Check List 

Tests 

 - Unit test
Passed
 - Integration test
 - No code

### DIFF
--- a/pump/storage/metrics.go
+++ b/pump/storage/metrics.go
@@ -38,7 +38,7 @@ var (
 		prometheus.CounterOpts{
 			Namespace: "binlog",
 			Subsystem: "pump_storage",
-			Name:      "deleted_kv",
+			Name:      "deleted_kv_total",
 			Help:      "deleted kv number",
 		})
 
--- a/pump/storage/metrics.go
+++ b/pump/storage/metrics.go
@@ -26,6 +26,22 @@ var (
 			Help:      "gc ts of storage",
 		})
 
+	doneGcTSGauge = prometheus.NewGauge(
+		prometheus.GaugeOpts{
+			Namespace: "binlog",
+			Subsystem: "pump_storage",
+			Name:      "done_gc_ts",
+			Help:      "the metadata and vlog after this gc ts has been collected",
+		})
+
+	deletedKv = prometheus.NewCounter(
+		prometheus.CounterOpts{
+			Namespace: "binlog",
+			Subsystem: "pump_storage",
+			Name:      "deleted_kv",
+			Help:      "deleted kv number",
+		})
+
 	storageSizeGauge = prometheus.NewGaugeVec(
 		prometheus.GaugeOpts{
 			Namespace: "binlog",
@@ -97,6 +113,8 @@ var (
 // InitMetircs register the metrics to registry
 func InitMetircs(registry *prometheus.Registry) {
 	registry.MustRegister(gcTSGauge)
+	registry.MustRegister(doneGcTSGauge)
+	registry.MustRegister(deletedKv)
 	registry.MustRegister(maxCommitTSGauge)
 	registry.MustRegister(tikvQueryCount)
 	registry.MustRegister(errorCount)
--- a/pump/storage/storage.go
+++ b/pump/storage/storage.go
@@ -703,8 +703,6 @@ func (a *Append) doGCTS(ts int64) {
 			}
 		}
 
-		iter.Release()
-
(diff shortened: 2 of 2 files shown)


july2993:
any more detail about why it will be so slow, will it be relate to because too many key with same 
 key encodeTSKey(0)(too many version) ?

july2993:
what about adding a metrics about deleted keys number?, so we can know the delete speed later from metrics.

pump/storage/storage.go
```diff
@@ -660,6 +660,12 @@ func (a *Append) doGCTS(ts int64) {
 
 	deleteNum := 0
 
+	irange := &util.Range{
+		Start: encodeTSKey(0),
+		Limit: encodeTSKey(ts + 1),
+	}
+	iter := a.metadata.NewIterator(irange, nil)
```
suzaku: In the original implementation, the `iter` is created and released every time in the loop to scan all the remaining records in the range `[0, ts]`.
After this change, we'll be using the `iter` after releasing it, which seems buggy?
To trigger the bug, we'll need to GC at least `1024 * 100` records, which explains why the tests were passing.

leoppro:
@july2993 @suzaku PTAL

leoppro:
> any more detail about why it will be so slow, will it be relate to because too many key with same
> key encodeTSKey(0)(too many version) ?
> 
> will `db.Get(key)` cost so long too?

I will go deep into this problem and find the answer to those.

pump/storage/metrics.go
```diff
@@ -26,6 +26,22 @@ var (
 			Help:      "gc ts of storage",
 		})
 
+	downGcTSGauge = prometheus.NewGauge(
```
july2993: s/down/done

pump/storage/metrics.go
```diff
@@ -26,6 +26,22 @@ var (
...
+		prometheus.GaugeOpts{
+			Namespace: "binlog",
+			Subsystem: "pump_storage",
+			Name:      "done_gc_ts",
+			Help:      "the metadata and vlog after this gc ts has been collected",
+		})
+
+	gcSpeed = prometheus.NewGauge(
```
july2993: you should add a Counter of deleted number instead of Gauge

pump/storage/storage.go
```diff
@@ -703,27 +703,30 @@ func (a *Append) doGCTS(ts int64) {
...
-		iter.Release()
-
 		if deleteBatch < 100 {
 			if batch.Len() > 0 {
 				err := a.metadata.Write(batch, nil)
 				if err != nil {
 					log.Error("write batch failed", zap.Error(err))
 				}
 				batch.Reset()
+				deletedKv.Add(float64(deleteBatch * 1024))
```
july2993: Add(batch.Len()) before Reset batch

pump/storage/storage.go
```diff
@@ -703,27 +703,30 @@ func (a *Append) doGCTS(ts int64) {
...
 			}
 			break
 		}
 
 		if len(lastKey) > 0 {
 			a.vlog.gcTS(decodeTSKey(lastKey))
 		}
-
+		deletedKv.Add(float64(deleteBatch * 1024))
```
july2993: ditto

leoppro:
@july2993 PTAL again

suzaku:
@leoppro You may consider changing the commit message to "Avoid repeated called of NewIterator which might be slow" before merging.

pump/storage/metrics.go
```diff
@@ -26,6 +26,22 @@ var (
...
+			Subsystem: "pump_storage",
+			Name:      "done_gc_ts",
+			Help:      "the metadata and vlog after this gc ts has been collected",
+		})
+
+	deletedKv = prometheus.NewCounter(
+		prometheus.CounterOpts{
+			Namespace: "binlog",
+			Subsystem: "pump_storage",
+			Name:      "deleted_kv",
```
july2993: s/deleted_kv/deleted_kv_total

leoppro:
```suggestion
			Name:      "deleted_kv_total",
```

PR #721, opened 2019-08-20
Change default cache count from 65536 to 512

lichunzhu:
A son PR of https://github.com/pingcap/tidb-binlog/pull/715.
Change default cache count from 65536 to workcount*batchsize
Set default cache count as a variable of config. The default size is changed from 65536 `workcount*batchsize`.

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

Side effects

Related changes

### DIFF
--- a/arbiter/server.go
+++ b/arbiter/server.go
@@ -182,10 +182,11 @@ func (s *Server) Run() error {
 
 	var syncErr error
 
+	syncCtx, syncCancel := context.WithCancel(ctx)
 	wg.Add(1)
 	go func() {
 		defer wg.Done()
-		syncErr = syncBinlogs(s.kafkaReader.Messages(), s.load)
+		syncErr = syncBinlogs(syncCtx, s.kafkaReader.Messages(), s.load)
 		if syncErr != nil {
 			s.Close()
 		}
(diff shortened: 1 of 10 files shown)


lichunzhu:
@suzaku So what do you think of the right value? How about 512? I'm not so sure about this.

suzaku:
@lichunzhu Note that `cache-binlog-size` can not avoid OOM in all cases, because the buffer is not the only data structure consuming memory. For example, if the downstream is TiDB/MySQL, we'll be exeucting SQLs in bulk and the binlogs not executed yet are stored in a slice which might take up a lot of memory if `worker-count` and `batch-size` is set to large numbers.

What I am suggesting is that, since we are facing a complicated problem with no perfect solution right now, we'd better just use a simple solution that can make OOM less likely.

lichunzhu:
@suzaku I understand what you mean now. Then what do you think of the limit value? Is 512 OK?

july2993:
@lichunzhu just change to 512 but don't add the config in `SyncerConfig`

suzaku:
512 or anything less than or equal to 1024 should be a safe choice, because in loader we also have a buffered channel which is hard coded to be 1024 in length.

lichunzhu:
@suzaku @july2993 I have modified `defaultBinlogItemCount` from 65536 to 512, PTAL.

suzaku:
@lichunzhu Just writing `512` may be more readable than `1 >> 9`?

lichunzhu:
> @lichunzhu Just writing `512` may be more readable than `1 >> 9`?

Revised.

PR #722, opened 2019-08-20
fix iter may not be Released in some cases

leoppro:
When error happened in loop,  iter may not be released. And when `l0Num >= l0Trigger`, we better release the iter, it may be affect compaction operation.

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

### DIFF
--- a/pump/storage/storage.go
+++ b/pump/storage/storage.go
@@ -34,6 +34,7 @@ import (
 	"github.com/pingcap/tidb/store/tikv/oracle"
 	pb "github.com/pingcap/tipb/go-binlog"
 	"github.com/syndtr/goleveldb/leveldb"
+	"github.com/syndtr/goleveldb/leveldb/iterator"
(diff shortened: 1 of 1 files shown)


leoppro:
@suzaku PTAL

pump/storage/storage.go
```diff
@@ -681,13 +688,24 @@ func (a *Append) doGCTS(ts int64) {
...
+			}
 			time.Sleep(5 * time.Second)
 			continue
 		}
 
 		deleteBatch := 0
 		var lastKey []byte
 
+		if iter == nil {
+			log.Info("create new gc iter", zap.Int64("ts", ts),
```
suzaku:
```suggestion
			log.Info("New LevelDB iterator created for GC", zap.Int64("ts", ts),
```

PR #723, opened 2019-08-21
add drainer notify test for pump node

lichunzhu:
[TOOL-1056](https://internal.pingcap.net/jira/browse/TOOL-1056)

Mock a fake grpc drainer to be notified. Mocked drainer will occupy port 8249 for a while.

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

Side effects

Related changes

### DIFF
--- a/pump/node_test.go
+++ b/pump/node_test.go
@@ -179,6 +179,7 @@ var _ = Suite(&notifyDrainerSuite{})
 type mockDrainerServer struct {
 	binlog.CisternServer
 	notified bool
+	host     string
 }
 
 func (s *mockDrainerServer) Notify(ctx context.Context, in *binlog.NotifyReq) (*binlog.NotifyResp, error) {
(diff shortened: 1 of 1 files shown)


pump/node_test.go
```diff
@@ -168,3 +171,48 @@ func (s *heartbeatSuite) TestShouldUpdateStatus(c *C) {
...
+}
+
+func (s *notifyDrainerSuite) TestNotifyDrainer(c *C) {
+	host := "127.0.0.1:8249"
+	lis, err := net.Listen("tcp", host)
+	c.Assert(err, IsNil)
+	gs := grpc.NewServer()
+	mockDrainerServerImpl := &mockDrainerServer{notified: false}
+	binlog.RegisterCisternServer(gs, mockDrainerServerImpl)
+	// defer gs.Stop()
```
suzaku: Why is this line commented?

pump/node_test.go
```diff
@@ -168,3 +171,48 @@ func (s *heartbeatSuite) TestShouldUpdateStatus(c *C) {
...
+		},
+	}
+	drainerNodeStatus := &node.Status{
+		NodeID: "pump_notify_test",
+		State:  node.Online,
+		Addr:   host,
+	}
+	err = registry.UpdateNode(context.Background(), "drainers", drainerNodeStatus)
+	c.Assert(err, IsNil)
+	err = pNode.Notify(context.Background())
```
suzaku: There's a basic structure in most tests:

1. Setup dependencies
1. Run the target against some input
1. Validate the output

We can arrange our test in three paragraphs and insert blank lines between them to make it clearer.

For each paragraph, we may add comment to explain what's being setup if it's not clear from the context.

In this case, the setup step is quite a lot of work, you may want to make it simpler:

1. Is it possible to extract some functions?
1. Is the etcd setup necessary? Can we replace it with a mock?

pump/node_test.go
lichunzhu: Revised.

pump/node_test.go
lichunzhu: 1. I simply the start of mock drainer as a function of `mockDrainerServer`.
2. IMO, the price of mock etcd setup is much higher than current implementation. Especially in function `pump/node.go:176:notifyDrainer` many varibles are very hard to mock.

july2993:
@leoppro PTAL

PR #726, opened 2019-08-23
pump: Clean up obsolete value logs concurrently

suzaku:

Deleting value log files before deleting the corresponding rows in
GoLevelDB doesn't affect correctness in this case, because:

1. the specified Ts is stored in LevelDB as gcTS before actually doing GC
2. PullCommitBinlogs uses `gcTS + 1` as the start of the range when iterating
   binlogs, so even if all the records in LevelDB are not deleted, we can
   still be sure that the garbage collected rows are not sent to drainers

So we don't have to wait for the potentially slow LevelDB cleanup to
remove unused value log files and reclaim most of the disk space.

1. Call `vlog.gcTS` in a separate goroutine.
2. Make sure `vlog.gcTS` doesn't remove files while they are being
    scanned.

### Check List 

Tests 

 - Unit test

Code changes

Side effects

Related changes


pump/storage/vlog.go
```diff
@@ -413,6 +424,10 @@ func (vlog *valueLog) scanRequests(start valuePointer, fn func(*request) error)
 //    that every single binlog added would be visited
 // 2. If GC is running concurrently, logFiles may be closed and deleted, thus breaking the scanning.
 func (vlog *valueLog) scan(start valuePointer, fn func(vp valuePointer, record *Record) error) error {
+	vlog.gcGuard <- struct{}{}
```
leoppro: why `scan` and `gcTS` should be mutually exclusive.
if we can guarantee that `start` is safe(which won't be removed), it's fine that `scan` and `gcTS` would be executed simultaneously

suzaku: But it seems we have nothing in the program to make sure the scanned range doesn't overlap with the GC range.

Note that a slow `scan` may be in progress, while `gcTS` is called, in the current implementation, both calls would consider `start` and `ts` safe.

pump/storage/storage.go
```diff
@@ -651,7 +650,14 @@ func (a *Append) GC(ts int64) {
 }
 
 func (a *Append) doGCTS(ts int64) {
-	log.Info("start gc", zap.Int64("ts", ts))
+	log.Info("Starting GC", zap.Int64("ts", ts))
+
+	var wg sync.WaitGroup
+	wg.Add(1)
+	go func() {
```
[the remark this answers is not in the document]
suzaku: Make sense.

pump/storage/vlog.go
leoppro: could we try to make sure that the scan operate is safe, but not to avoid conflicts.

in another way, `scan` is a time-consuming and frequent operation, however `vlog gc` will be triggered hourly. it seems that it has a high probability of running `scan` when `vlog gc` is triggered. we still can't free up disk effectively.

when `done_gc_tso` too far behind `gc_tso`, once `gc` process could be continued a few weeks, if `scan` is running, unfortunately, `vlog gc` will be triggered after a few weeks.

suzaku: `vlog.scan` is only called in two places:

1. When the storage is starting up
1. When the slow chaser tries to catch up from the last unread binlog

These are not frequent operations.

leoppro: fine, but I still suggest adding a retry mechanism for `vlog gc`

pump/storage/storage.go
```diff
@@ -651,7 +650,9 @@ func (a *Append) GC(ts int64) {
 }
 
 func (a *Append) doGCTS(ts int64) {
-	log.Info("start gc", zap.Int64("ts", ts))
+	log.Info("Starting GC", zap.Int64("ts", ts))
+
+	a.vlog.gcTS(ts)
```
lichunzhu: May I ask a question here? If we `gcTS(ts)` here, why we need to do `a.vlog.gcTS(decodeTSKey(lastKey))`? Because the storage is concurrently adding vlogs?

leoppro: yes, we store metadata( ts -> (vlogfile,offset) ) in goleveldb, and store binlog data in vlog.

suzaku: @lichunzhu Good observation! I think we can get rid of `a.vlog.gcTS(decodeTSKey(lastKey))` if we are doing the whole `gcTS` thing concurrently.

pump/storage/vlog.go
suzaku: Updated

leoppro:
why `DNM`

july2993:
> why `DNM`

I will take a look later.

suzaku:

pump/storage/storage.go
```diff
@@ -736,15 +742,13 @@ func (a *Append) doGCTS(ts int64) {
...
 			doneGcTSGauge.Set(float64(oracle.ExtractPhysical(uint64(decodeTSKey(lastKey)))))
 		}
 		log.Info("has delete", zap.Int("delete num", deleteNum))
 	}
-
-	a.vlog.gcTS(ts)
+	wg.Wait()
 	doneGcTSGauge.Set(float64(oracle.ExtractPhysical(uint64(ts))))
-	log.Info("finish gc", zap.Int64("ts", ts), zap.Int("delete num", deleteNum))
+	log.Info("Finish GC", zap.Int64("ts", ts), zap.Int("delete num", deleteNum))
```
GregoryIan: it's bad to output this log if we break from `for loop`, it will make debug not clear enough

GregoryIan:
the LGTM is too late, Good Job!

GregoryIan:
LGTM

PR #728, opened 2019-08-26
test: add test for metadata gc 

leoppro:
we don't have a test case when the total number of kv is large.

I just turn up the total number of kv and gc ts.
and i fix a bug by the way,it will skip a key by mistake when `deleteBatch = 100`.

### Check List 

Tests 

 - Unit test
 - Integration test

### DIFF
--- a/pump/storage/storage_test.go
+++ b/pump/storage/storage_test.go
@@ -277,7 +277,7 @@ func (as *AppendSuit) TestDoGCTS(c *check.C) {
 	defer cleanAppend(append)
 
 	var i int64
-	var n int64 = 1024 * 500
+	var n int64 = 1025 * 500
 	var batch leveldb.Batch
 	for i = 1; i < n; i++ {
 		batch.Put(encodeTSKey(i), value)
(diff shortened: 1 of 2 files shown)


leoppro:
@july2993 @lichunzhu PTAL

pump/storage/storage_test.go
```diff
@@ -277,15 +277,15 @@ func (as *AppendSuit) TestDoGCTS(c *check.C) {
 	defer cleanAppend(append)
 
 	var i int64
-	var n int64 = 1 << 11
+	var n int64 = 1024 * 500
```
lichunzhu: I think we can use `1024 * 500 + 500` rather than `1024 * 500` to test branch of line 727 in `storage.go`.

leoppro:
```suggestion
	var n int64 = 1023 * 500
```

leoppro:
PTAL again @lichunzhu

lichunzhu:
LGTM

PR #729, opened 2019-08-26
Avoid risk of ignoring failed to read checkpoint

lichunzhu:
[TOOL-1520](https://internal.pingcap.net/jira/browse/TOOL-1520)
In previous `drainer/checkpoint/mysql.go:Load()` we use `query` to get the checkpoint data but never check the `row.Err()` which may skip some failures but the `checkpoint` is still updated.
Modify `Load()` in mysql checkpoint. Current version will use `queryrow` instead of `query` to avoid missing some errors.

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

Side effects

Related changes

### DIFF
--- a/drainer/checkpoint/mysql.go
+++ b/drainer/checkpoint/mysql.go
@@ -94,7 +94,6 @@ func (sp *MysqlCheckPoint) Load() error {
 		sp.CommitTS = sp.initialCommitTS
 		return nil
 	case err != nil:
-		log.Errorf("select checkPoint error %v", err)
 		return errors.Annotatef(err, "QueryRow failed, sql: %s", selectSQL)
 	}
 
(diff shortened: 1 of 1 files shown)


july2993:
please update the description, don't just show a not-public issue

drainer/checkpoint/mysql.go
```diff
@@ -86,29 +86,19 @@ func (sp *MysqlCheckPoint) Load() error {
...
-
-	if len(str) == 0 {
+	selectSQL := genSelectSQL(sp)
+	err := sp.db.QueryRow(selectSQL).Scan(&str)
+	switch {
+	case err == sql.ErrNoRows:
 		sp.CommitTS = sp.initialCommitTS
 		return nil
+	case err != nil:
+		log.Errorf("select checkPoint error %v", err)
```
july2993: drop this

lichunzhu: Revised

lichunzhu:
> please update the description, don't just show a not-public issue

Updated.

PR #735, opened 2019-08-30
drainer/: Change unreasonable buff size and reduce memory usage

july2993:
the `maxBinlogItemCount` var will affect the job buffer size in an unreasonable way.

In syncer.sync will work like this

```
// where will be worker-count goroutine to run this
for {
    get job from "jobChan chan *job"
    if reach some condition {
          execute the pending job(execute sql)
     } else {
         continue
     }
}
```
note execute the sql may take times, and not consume the job chan.
so the upstream may block at add job into the chan, and also will not add job the other `sync` job chan.

avoid any buffer at upstream  to much mem and not decrease performance, and don't make buffer size of job chan depent on `maxBinlogItemCount`

Screen Shot 2019-08-30 at 2 43 43 PM

the flowing v1 means https://github.com/pingcap/tidb-binlog/pull/735/commits/f9e4589de8b934f3dcfcdee097fd3c9206d3d459
v2 means one more commit https://github.com/pingcap/tidb-binlog/pull/735/commits/3074b3bb6b9e526eccbb036377a94473b167df5e

before 10:53   version: v2.1.16 maxBinlogItemCount=16
start at 10:53   version: v1  maxBinlogItemCount=16
start at 30 11:17   version:v2.1.16   maxBinlogItemCount=64k (the previous default value)
start at 30 11:47   version: v1 maxBinlogItemCount=0

all using worker-count = 256, txn-batch = 20
note when the drainer restarts, the first 5 minutes will use safe-mode, the event metrics count the changed events.

## Another test:
the upstream data one binlog near 60M

start at 19:25: version: v1
start at 19:45: version: v2
Screen Shot 2019-08-30 at 7 56 23 PM
Screen Shot 2019-08-30 at 7 56 14 PM

### Check List 

Tests 

 - Unit test
 - Integration test

### DIFF
--- a/drainer/checkpoint/mysql.go
+++ b/drainer/checkpoint/mysql.go
@@ -92,6 +92,7 @@ func (sp *MysqlCheckPoint) Load() error {
 	switch {
 	case err == sql.ErrNoRows:
 		sp.CommitTS = sp.initialCommitTS
+		log.Info("no checkpoint in downstream table, use %d", sp.initialCommitTS)
 		return nil
 	case err != nil:
 		return errors.Annotatef(err, "QueryRow failed, sql: %s", selectSQL)
@@ -100,6 +101,8 @@ func (sp *MysqlCheckPoint) Load() error {
 	if err = json.Unmarshal([]byte(str), sp); err != nil {
 		return errors.Trace(err)
 	}
+
+	log.Info("read checkpoint from downstream table: %s", sp.CommitTS)
 
 	return nil
 }
--- a/drainer/checkpoint/mysql.go
+++ b/drainer/checkpoint/mysql.go
@@ -92,7 +92,7 @@ func (sp *MysqlCheckPoint) Load() error {
 	switch {
 	case err == sql.ErrNoRows:
 		sp.CommitTS = sp.initialCommitTS
-		log.Info("no checkpoint in downstream table, use %d", sp.initialCommitTS)
+		log.Infof("no checkpoint in downstream table, use initialCommitTS: %d", sp.initialCommitTS)
 		return nil
 	case err != nil:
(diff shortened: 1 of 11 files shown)


GregoryIan:
lgtm

leoppro:
why ci is failed?

july2993:
> why ci was failed?

i don't know, failed at `status` test, but not related to this pr.
(run ok locally and failed ci at another pr too-2.1)

leoppro:
LGTM

lichunzhu:
IMO, CI fails because drainer hasn't consumed the pump 8250's storaged binlogs in 15 seconds when closing pump. This problem can be replayed on my server.

july2993:
> IMO, CI fails because drainer hasn't consumed the pump 8250's storaged binlogs in 15 seconds when closing pump. This problem can be reproduced on my server.

Is it something relate to this pr change? can you fix this?

lichunzhu:
> > IMO, CI fails because drainer hasn't consumed the pump 8250's storaged binlogs in 15 seconds when closing pump. This problem can be reproduced on my server.
> 
> Is it something relate to this pr change? can you fix this?

I'm not so sure now. I'm working on that now and later I will get a conclusion.

lichunzhu:
There is still 4s gap between drainer and pump after 15s. I guess when drainer has cache it can cache some binlogs from pump then drainer can still sort and execute the cached binlogs even though this pump is paused. Maybe `binlogChanSize` shouldn't be set to 0. Or we can set check time in `check_status` more bigger.

july2993:
@lichunzhu PTAL, the root cause is when runs into the `status` test, it will start with commit-ts = 0 and fetch data from the oldest point.

PR #736, opened 2019-09-02
storage: add a test for storage gc

leoppro:
Our test cases didn't cover `if l0Num >= l0Trigger{}` in `doGCTS` function

I import `failpoint` to inject a big `l0Num`.

### Check List 

Tests 

 - Unit test
 - Integration test

### DIFF
--- a/pump/storage/storage.go
+++ b/pump/storage/storage.go
@@ -790,7 +790,6 @@ func (a *Append) batchGC(startTs int64, endTs int64, batchSize int, l0Trigger in
 		}
 		log.Info("has delete", zap.Int("delete num", deleteNum))
 	}
-	return
 }
 
 // MaxCommitTS implement Storage.MaxCommitTS
--- a/pump/storage/storage.go
+++ b/pump/storage/storage.go
@@ -689,21 +689,27 @@ func (a *Append) doGCTS(ts int64) {
 	if a.options.KVConfig != nil && a.options.KVConfig.CompactionL0Trigger > 0 {
 		l0Trigger = a.options.KVConfig.CompactionL0Trigger
 	}
+
+	getL0Num := func() (int, error) {
+		nStr, err := a.metadata.GetProperty("leveldb.num-files-at-level0")
+		if err != nil {
+			return 0, errors.Annotate(err, "get `leveldb.num-files-at-level0` property failed")
+		}
+
+		l0Num, err := strconv.Atoi(nStr)
+		if err != nil {
+			return 0, errors.Annotatef(err, "parse `leveldb.num-files-at-level0` result to int failed, str: %s", nStr)
+		}
+		return l0Num, nil
+	}
+
 	var alreadyGcTS int64
 	for {
-		alreadyGcTS = a.batchGC(alreadyGcTS, ts, 100, l0Trigger, func() (int, error) {
-			nStr, err := a.metadata.GetProperty("leveldb.num-files-at-level0")
-			if err != nil {
-				return 0, errors.Annotate(err, "get `leveldb.num-files-at-level0` property failed")
-			}
-
-			l0Num, err := strconv.Atoi(nStr)
-			if err != nil {
-				return 0, errors.Annotatef(err, "parse `leveldb.num-files-at-level0` result to int failed, str: %s", nStr)
-			}
-			return l0Num, nil
-		})
-
+		alreadyGcTS, err := a.batchGC(alreadyGcTS, ts, 100, l0Trigger, getL0Num)
+		if err != nil {
+			log.Error("batch GC failed", zap.Error(err))
+			break
+		}
 		if alreadyGcTS >= ts {
 			break
 		}
@@ -711,10 +717,9 @@ func (a *Append) doGCTS(ts int64) {
 	}
 
 	wg.Wait()
-	doneGcTSGauge.Set(float64(oracle.ExtractPhysical(uint64(ts))))
 }
 
-func (a *Append) batchGC(startTS int64, endTS int64, batchSize int, l0Trigger int, getL0Num func() (int, error)) (alreadyGcTS int64) {
+func (a *Append) batchGC(startTS int64, endTS int64, batchSize int, l0Trigger int, getL0Num func() (int, error)) (alreadyGcTS int64, err error) {
 
 	batch := new(leveldb.Batch)
 	deleteNum := 0
@@ -727,9 +732,7 @@ func (a *Append) batchGC(startTS int64, endTS int64, batchSize int, l0Trigger in
 
 	log.Info("New LevelDB iterator created for GC",
 		zap.Int64("startTS", startTS),
-		zap.Int64("endTS", endTS),
-		zap.Int64("start", decodeTSKey(irange.Start)),
-		zap.Int64("limit", decodeTSKey(irange.Limit)))
+		zap.Int64("endTS", endTS))
 	iter := a.metadata.NewIterator(irange, nil)
 	defer func() {
 		if iter != nil {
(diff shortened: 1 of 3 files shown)


leoppro:
PTAL @suzaku @lichunzhu

pump/storage/storage.go
```diff
@@ -673,6 +674,7 @@ func (a *Append) doGCTS(ts int64) {
 		}
 	}()
 
+	l0TestLarge := 0
 	for {
 		nStr, err := a.metadata.GetProperty("leveldb.num-files-at-level0")
```
suzaku: Maybe just extract a function to check if further writes are going to cause compaction? Then we can simply mock the function in our test without introducing failpoint?

pump/storage/storage_test.go
```diff
@@ -272,20 +273,21 @@ func (as *AppendSuit) testWriteBinlogAndPullBack(c *check.C, prewriteValueSize i
...
 
 func (as *AppendSuit) TestDoGCTS(c *check.C) {
+	failpoint.Enable("github.com/pingcap/tidb-binlog/pump/storage/l0-file-too-much", "return")
 	var value = make([]byte, 10)
 	append := newAppend(c)
 	defer cleanAppend(append)
 
 	var i int64
-	var n int64 = 1025 * 500
+	var n int64 = 1026 * 3214
```
suzaku: If the limit per batch is configurable, we can use some smaller value here so that the tests don't slow down.

pump/storage/storage.go
leoppro: I will use failpoint to inject error into `fsync`, here we intorduced failpoint is prepare for `fsync` test.

pump/storage/storage_test.go
leoppro: PTAL again @suzaku

pump/storage/storage.go
```diff
@@ -686,6 +688,15 @@ func (a *Append) doGCTS(ts int64) {
...
 		}
 
+		failpoint.Inject("l0-file-too-much", func() {
+			if l0TestLarge < l0Trigger {
+				l0TestLarge++
+			} else {
+				l0TestLarge -= l0Trigger
+			}
+			l0Num = l0TestLarge
+		})
```
suzaku: Can you explain what cases are setup here?

leoppro:
@suzaku PTAL again

pump/storage/storage.go
suzaku: Please add test to check that the waiting happens when the number of l0 files is larger than l0Trigger.
And mock the wait time below to make it shorter.

pump/storage/storage.go
```diff
@@ -686,12 +688,22 @@ func (a *Append) doGCTS(ts int64) {
...
+
 		if l0Num >= l0Trigger {
 			log.Info("wait some time to gc cause too many L0 file", zap.Int("files", l0Num))
 			if iter != nil {
 				iter.Release()
 				iter = nil
 			}
+			failpoint.Continue()
```
leoppro: here, when we switch on the failpoint, this line will be rewritten to `continue`, so we skip the `Sleep` in test. @suzaku

suzaku: I see. But we still need to validate that the behavior "LevelDB GC doesn't happen when there are too many l0 files".

pump/storage/storage.go
```diff
@@ -674,28 +674,13 @@ func (a *Append) doGCTS(ts int64) {
 	}()
 
 	for {
-		nStr, err := a.metadata.GetProperty("leveldb.num-files-at-level0")
+		compacted, err := waitL0FilesCompacted(a.metadata, l0Trigger, 5*time.Second, &iter)
```
suzaku: In my opinion, having to pass the `&iter` reference just for cleaning up means we are extracting the wrong function. Ideally, the extracted function should know little about the original function so that it could later be reused in other cases (for example, where there's no Iterator involved).

pump/storage/storage_test.go
```diff
@@ -298,6 +299,23 @@ func (as *AppendSuit) TestDoGCTS(c *check.C) {
 	}
 }
 
+func (as AppendSuit) TestWaitL0FilesCompacted(c *check.C) {
```
july2993: this can't test `l0Num >= l0Trigger` -> l0Num < l0Trigger, it's a little confusing about the name  `TestWaitL0FilesCompacted`

leoppro: https://github.com/pingcap/tidb-binlog/pull/736/commits/2b07c08baf67a47d19a8d2f6d2e1f19fbfb6041c

what about this kind of implementation which use failpoint?

leoppro: https://github.com/pingcap/tidb-binlog/pull/736/commits/2b07c08baf67a47d19a8d2f6d2e1f19fbfb6041c

what about this kind of implementation which use failpoint?

pump/storage/storage.go
```diff
@@ -756,6 +741,31 @@ func isFakeBinlog(binlog *pb.Binlog) bool {
...
+	l0Num, err := strconv.Atoi(nStr)
+	if err != nil {
+		log.Error("parse `leveldb.num-files-at-level0` result to int failed", zap.String("str", nStr), zap.Error(err))
+		return false, err
+	}
+
+	if l0Num >= l0Trigger {
+		log.Info("wait some time to gc cause too many L0 file", zap.Int("files", l0Num))
+		if *iter != nil {
+			(*iter).Release()
```
GregoryIan: I don't suggest to put a `iterator` object into two function scope

leoppro:
@suzaku @july2993 @GregoryIan  PTAL again

pump/storage/storage.go
```diff
@@ -707,39 +739,21 @@ func (a *Append) doGCTS(ts int64) {
...
-			log.Error("get `leveldb.num-files-at-level0` property failed", zap.Error(err))
-			break
-		}
-
-		l0Num, err := strconv.Atoi(nStr)
+		l0Num, err := getL0Num()
 		if err != nil {
-			log.Error("parse `leveldb.num-files-at-level0` result to int failed", zap.String("str", nStr), zap.Error(err))
-			break
+			log.Error("get `l0Num` property of leveldb failed", zap.Error(err))
```
GregoryIan:
```suggestion
			log.Error("", zap.Error(err))
```

GregoryIan: messages are already in L697 and L702

pump/storage/storage.go
```diff
@@ -686,19 +685,52 @@ func (a *Append) doGCTS(ts int64) {
...
+			break
+		}
+		time.Sleep(5 * time.Second)
+	}
+
+	wg.Wait()
+	doneGcTSGauge.Set(float64(oracle.ExtractPhysical(uint64(ts))))
+}
+
+func (a *Append) batchGC(startTs int64, endTs int64, batchSize int, l0Trigger int, getL0Num func() (int, error)) (alreadyGcTS int64) {
```
GregoryIan:
```suggestion
func (a *Append) batchGC(startTS int64, endTS int64, batchSize int, l0Trigger int, getL0Num func() (int, error)) (alreadyGcTS int64) {
```

`alreadyGcTS ` and `startTs`, please try to keep the same style

pump/storage/storage.go
```diff
@@ -707,39 +739,21 @@ func (a *Append) doGCTS(ts int64) {
...
-			log.Error("get `leveldb.num-files-at-level0` property failed", zap.Error(err))
-			break
-		}
-
-		l0Num, err := strconv.Atoi(nStr)
+		l0Num, err := getL0Num()
 		if err != nil {
-			log.Error("parse `leveldb.num-files-at-level0` result to int failed", zap.String("str", nStr), zap.Error(err))
-			break
+			log.Error("", zap.Error(err))
```
suzaku: Empty message?

pump/storage/storage.go
```diff
@@ -686,19 +685,52 @@ func (a *Append) doGCTS(ts int64) {
...
+			if err != nil {
+				return 0, errors.Annotate(err, "get `leveldb.num-files-at-level0` property failed")
+			}
 
+			l0Num, err := strconv.Atoi(nStr)
+			if err != nil {
+				return 0, errors.Annotatef(err, "parse `leveldb.num-files-at-level0` result to int failed, str: %s", nStr)
+			}
+			return l0Num, nil
+		})
```
suzaku: Assign this function to a variable outside the loop and use the variable here would make the code clearer?

pump/storage/storage.go
```diff
@@ -707,39 +739,21 @@ func (a *Append) doGCTS(ts int64) {
...
+		l0Num, err := getL0Num()
 		if err != nil {
-			log.Error("parse `leveldb.num-files-at-level0` result to int failed", zap.String("str", nStr), zap.Error(err))
-			break
+			log.Error("", zap.Error(err))
+			return
 		}
 
 		if l0Num >= l0Trigger {
 			log.Info("wait some time to gc cause too many L0 file", zap.Int("files", l0Num))
```
suzaku: Now that this is an independent function, this log message should be updated.

pump/storage/storage.go
```diff
@@ -764,18 +778,18 @@ func (a *Append) doGCTS(ts int64) {
 				deletedKv.Add(float64(batch.Len()))
 				batch.Reset()
 			}
-			log.Info("Finish KV GC", zap.Int64("ts", ts), zap.Int("delete num", deleteNum))
-			break
+			alreadyGcTS = endTS
```
suzaku: This line is a little tricky, although this is extracted as an independent function, this line assumes that the returned value is used in the `alreadyGcTS >= ts` check. If someone later uses the returned value and expects it to be the real last ts, he would be wrong.

pump/storage/storage.go
leoppro: It's @GregoryIan 's suggest.
He thinks that messages are already in L697 and L702

pump/storage/storage.go
leoppro: I agree with you

what do you think about that @GregoryIan

pump/storage/storage.go
GregoryIan: why not put L787 ～ L790 before L772？

pump/storage/storage.go
GregoryIan: which log message should updated to？

pump/storage/storage.go
```diff
@@ -686,19 +685,52 @@ func (a *Append) doGCTS(ts int64) {
...
+			if err != nil {
+				return 0, errors.Annotatef(err, "parse `leveldb.num-files-at-level0` result to int failed, str: %s", nStr)
+			}
+			return l0Num, nil
+		})
+
+		if alreadyGcTS >= ts {
+			break
+		}
+		time.Sleep(5 * time.Second)
```
GregoryIan: the old code would break from `for` loop and then return at L719、L713, but now we change this logic, it would always retry if we fail to get `leveldb.num-files-at-level0`

pump/storage/storage.go
leoppro: the mean of `alreadyGcTS` is that the KVs before this TS are already removed. and it's not mean that this TS is the last one that was removed. 
this line `alreadyGcTS = endTS` is correct, because this function can promise that all KVs smaller than `alreadyGcTS` will be deleted

leoppro:
PTAL again @GregoryIan @suzaku @july2993

pump/storage/storage.go
suzaku: It's correct, but tricky. You might just explicitly add another return value to dictate to the caller whether there's more to be deleted.

pump/storage/storage.go
```diff
@@ -738,58 +741,37 @@ func (a *Append) batchGC(startTS int64, endTS int64, batchSize int, l0Trigger in
...
-				deleteBatch++
+		for deleteBatch := 0; deleteBatch < batchSize && hasNext; deleteBatch++ {
+			var lastKey []byte
+			for batch.Reset(); batch.Len() < 1024 && iter.Next(); deleteNum++ {
+				batch.Delete(iter.Key())
+				lastKey = iter.Key()
 			}
-		}
-
-		if deleteBatch < batchSize {
```
suzaku: Why not use this to check for `hasNext`?

pump/storage/storage.go
```diff
@@ -686,96 +685,95 @@ func (a *Append) doGCTS(ts int64) {
...
+	wg.Wait()
+}
+
+func (a *Append) batchGC(startTS int64, endTS int64, batchSize int, l0Trigger int, getL0Num func() (int, error)) (int64, bool, error) {
+
+	batch := new(leveldb.Batch)
 	deleteNum := 0
+	alreadyGcTS := startTS
 
 	irange := &util.Range{
```
GregoryIan: maybe we also should avoid this issue https://github.com/pingcap/tidb-binlog/pull/758

PR #737, opened 2019-09-02
drainer/: Reduce memory usage (#735)

lichunzhu:
Cherry pick of https://github.com/pingcap/tidb-binlog/pull/735.
[TOOL-1480](https://internal.pingcap.net/jira/browse/TOOL-1480)
Current drainer uses buffed channel to cache some instances which may cause OOM when facing  binlogs which consumes too much memory.
1. Avoid any buffer at upstream to avoid too much memory usage and not decreasing performance.
2. Add `txnManager` in loader to manage cached txn memory usage. Without `txnManager` the performance of loader will be affected.
### To check the effectiveness, we make two tests
In the following:
v1 means https://github.com/pingcap/tidb-binlog/commit/097db4237f5b64718cda5483243b8975b93041d2, which is the newest master branch.
v2 means https://github.com/pingcap/tidb-binlog/pull/737/commits/1e51514fd313f3b9783bd365b5f3cee3f8b78d5b, which is the current branch.
In v1, `maxBinlogItemCount=512`
In v2, `maxBinlogItemCount=0`
#### TEST 1: Small binlogs, to validate the efficiency of drainer
I put the test of v1 and v2 in one picture. 15:30 ~ 15:50 is test of v1 while 15:53 ~ 16:13 is test of v2. As we can see the memory usage and drainer event of both branches are similar.
drainer event:
image
memory:
image
execution time:
image

#### TEST 2: the upstream data one binlog near 60M, to validate the momery usage of drainer
For v1 and situations we set `loader.input`,`loader.success`, `dsyncer.success` buffer to 1024 drainer will get an OOM problem. The memory usage will rise up to more than 100G.
For v2, the memory usage is around 2G. The drainer event for DDL is around 2k and is around 14k for DMLs.
drainer event:
image
memory:
image
execute time:
image

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

Side effects

Related changes


lichunzhu:
DNM, test is in progress.

WangXiangUSTC:
LGTM

drainer/pump_test.go
```diff
@@ -80,10 +80,12 @@ func (s *pumpSuite) TestPullBinlog(c *C) {
 
 	// ascending commitTs order
 	pullBinlogCommitTSChecker(commitTsArray, ret, binlogBytesChan, c)
+	time.Sleep(10 * time.Microsecond)
```
suzaku: Why sleep here?

lichunzhu: code: https://github.com/pingcap/tidb-binlog/blob/master/drainer/pump.go#L173
Because when `pullBinlogCommitTSChecker` is finished and quits, `p.latestTS` has not been updated yet.Because`binlogChanSize = 0`, `pullBinlogCommitTSChecker` send out the final binlog and quit but in pump it needs some time to update `p.latestTS`. So we sleep here to assure that `p.latestTS` has been updated before `assert`.

lichunzhu:
DNM now. This PR should better be merged after release-2.1 version.

july2993:
do you run it at the case there lag ? 
In v1, maxBinlogItemCount=512

for this, the mem will not take only less than 1G

lichunzhu:
@july2993 Test results have been updated. PTAL.

lichunzhu:
Besides, I find that the main factor that affects the execution time is size of `loader.input` rather than `syncer.input`.

july2993:
`execution time` here only means how long takes to put into to chan.
you should focus on the event metrics.

In test1 the update drop from 7~8k to be below 1k, do you think this is acceptable?

lichunzhu:
I think that's not acceptable. I find that if I set size of `loader.input` to 8 the update can be around 2k. Is there any other solution?

lichunzhu:

lichunzhu:
@july2993 I have added txn manager in loader and the performance is not affected in this branch. The test results have been updated in the PR description. PTAL.

pkg/loader/load.go
```diff
@@ -430,16 +431,19 @@ func (s *loaderImpl) execDMLs(dmls []*DML) error {
...
 		close(s.successTxn)
+		txnManager.Close()
 	}()
 
 	batch := fNewBatchManager(s)
+	input := txnManager.put(s.input)
 
 	for {
 		select {
-		case txn, ok := <-s.input:
```
july2993: will it slow if only change s.input to be no buffer without using `txnManager` here ?

lichunzhu: image
Yes. From 10:27 on I set `loader.input` a 8 size buffer and don't use `txnManager` here. The Update Event drop from 7k+ to 3k+.

drainer/sync/mysql.go
```diff
@@ -107,7 +107,15 @@ func (m *MysqlSyncer) run() {
 
 		for txn := range m.loader.Successes() {
 			item := txn.Metadata.(*Item)
+
+			// clean useless item value to reduce memory usage
```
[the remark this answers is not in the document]
lichunzhu: image
From start time to 17:05 is the original branch. From 17:05 to the end time is set success chan buffer to 8. I find that reduce `success chan` lenth to 8 will not affect the sync speed.

pkg/loader/load.go
```diff
@@ -213,6 +213,7 @@ func (s *loaderImpl) markSuccess(txns ...*Txn) {
 		s.lastUpdateAppliedTSTime = time.Now()
 	}
 	for _, txn := range txns {
+		txn.DMLs = nil
```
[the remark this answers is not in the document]
lichunzhu: This part has been removed.

july2993:
@suzaku PTAL

pkg/loader/load.go
```diff
@@ -430,16 +430,19 @@ func (s *loaderImpl) execDMLs(dmls []*DML) error {
 
 // Run will quit when meet any error, or all the txn are drained
 func (s *loaderImpl) Run() error {
+	txnManager := newTxnManager(1024)
```
suzaku: 1. We can't control the memory consumed by the buffers here even when the capacity is limited by the number of DMLs.
1. If we do want to limit memory consumption by the number of DMLs, we can do that by replacing `chan *Txn` with `chan X` where X only contains single DML or DDL, we don't need to implement something similar to a buffered channel.

lichunzhu: 1. It's not easy to measure the memory consumed by a single DML. It seems that the main memory usage in one single DML is map `OldValues` and `Values`. So maybe we can count the sizeof these two maps?
2. If we use X contains only single DML or DDL, we also need to add `txn` to `batchManager` but not single DML or DDL. A successfully executed single DML doesn't mean the whole `txn` is successful.

pkg/loader/load.go
```diff
@@ -624,6 +629,82 @@ func (b *batchManager) put(txn *Txn) error {
...
+
+func (t *txnManager) pop(txn *Txn) {
+	t.cond.L.Lock()
+	t.cachedSize -= len(txn.DMLs)
+	t.cond.Signal()
+	t.cond.L.Unlock()
+}
+
+func (t *txnManager) Close() {
+	if t.closed {
```
GregoryIan: should we use atomic operation？

pkg/loader/load.go
suzaku: We can't avoid OOM by checking everywhere we might use memory (and making the program more complex), so the problem we need to solve is to lower the possibility, which I think can be solved by just setting smaller buffer sizes.

pkg/loader/load.go
lichunzhu: There won't be concurrent operations to change `t.closed` so I think we don't need to use atomic operation.

pkg/loader/load.go
```diff
@@ -624,6 +629,82 @@ func (b *batchManager) put(txn *Txn) error {
 	return nil
 }
 
+type txnManager struct {
```
GregoryIan: any test for this object?

pkg/loader/load.go
lichunzhu: The problem is that setting smaller buffer sizes will cause performance digression. If we just set buffer size of loader.input to 8 the Update Event wil drop from 7k+ to 3k+. 
image
If we change `default` in [loader](https://github.com/pingcap/tidb-binlog/blob/9f3c81683bb3428c4940611a6203288474d4aff0/pkg/loader/load.go#L456) to `case <-time.After(10 * time.Microsecond)` the situation will get better, but we still have performance digression. However, if the binlogs are very few, it takes more time for `load` to execute cached binlogs.
image

pkg/loader/load.go
lichunzhu: If in the end we decided to use `txnManager`, I will add some tests.

pkg/loader/load.go
GregoryIan: any better suggestions for @lichunzhu  ? @suzaku

pkg/loader/load.go
GregoryIan: YOU NEED TO ADD A COMMENT `IT'S NOT THREAD SAFE`

pkg/loader/load.go
```diff
@@ -624,6 +629,82 @@ func (b *batchManager) put(txn *Txn) error {
...
+func (t *txnManager) put(input chan *Txn) chan *Txn {
+	ret := t.cacheChan
+	go func() {
+		for !t.closed {
+			var txn *Txn
+			var ok bool
+			select {
+			case txn, ok = <-input:
+				if !ok {
+					close(ret)
```
GregoryIan: only close `ret` when `input` is closed?

GregoryIan:
This way is good, its essence is the token limit, in fact it be written more easily.

lichunzhu:
@GregoryIan I refine the code of `txnManager`. What's more, more comments and logs have been added. PTAL

pkg/loader/load.go
```diff
@@ -697,12 +707,13 @@ func (t *txnManager) pop(txn *Txn) {
 }
 
 func (t *txnManager) Close() {
-	if t.closed {
+	if atomic.LoadInt32(&t.isClosed) == 1 {
```
GregoryIan: use `CompareAndSwap`

lichunzhu: Revised.

pkg/loader/load.go
lichunzhu: Function `Close()` can close `ret` now.

pkg/loader/load.go
suzaku: Could you please change `output:   make(chan MergeItem)` back to `output:   make(chan MergeItem, 10)` and try again?

lichunzhu: @suzaku 
It still has performance digression. It's the situation which we set `default` to `case <-time.After(10 * time.Microsecond)`.
image

GregoryIan:
LGTM

GregoryIan:
@suzaku PTAL

lichunzhu:
DNM now, I will add some unit tests about `txnManager`.

GregoryIan:
please complete it today @lichunzhu

lichunzhu:

lichunzhu:
@GregoryIan PTAL

pkg/loader/load_test.go
```diff
@@ -378,6 +378,141 @@ func (s *batchManagerSuite) TestShouldExecAccumulatedDMLs(c *check.C) {
...
+	}
+	txn := &Txn{DMLs: dmls}
+	txnManager := newTxnManager(14, input)
+	output := txnManager.run()
+	// send 5 txns (size 3) to txnManager, the 5th txn should get blocked at cond.Wait()
+	for i := 0; i < 5; i++ {
+		select {
+		case input <- txn:
+		case <-time.After(50 * time.Microsecond):
+			c.Fatal("txnManager gets blocked while receiving txns")
```
suzaku: If the 5th txn get blocked sending to `input`, why didn't this test fail?

pkg/loader/load_test.go
```diff
@@ -378,6 +378,141 @@ func (s *batchManagerSuite) TestShouldExecAccumulatedDMLs(c *check.C) {
...
+		}
+	}
+	c.Assert(output, check.HasLen, 4)
+	// Next txn should be blocked
+	select {
+	case input <- txn:
+		c.Fatal("txnManager doesn't block the txn when room is not enough")
+	case <-time.After(50 * time.Microsecond):
```
suzaku:
```suggestion
	default:
```

pkg/loader/load_test.go
```diff
@@ -378,6 +378,141 @@ func (s *batchManagerSuite) TestShouldExecAccumulatedDMLs(c *check.C) {
...
+	}
+	c.Assert(output, check.HasLen, 4)
+	// pick one txn from output channel
+	select {
+	case t := <-output:
+		txnManager.pop(t)
+		c.Assert(t, check.DeepEquals, txn)
+	case <-time.After(50 * time.Microsecond):
```
suzaku: No need to wait? Just use `default`, because if we can't get a txn at this point, something must be wrong.

pkg/loader/load_test.go
```diff
@@ -378,6 +378,141 @@ func (s *batchManagerSuite) TestShouldExecAccumulatedDMLs(c *check.C) {
...
+	txnBig := &Txn{DMLs: []*DML{
+		{Tp: UpdateDMLType},
+		{Tp: InsertDMLType},
+		{Tp: UpdateDMLType},
+	}}
+	txnManager := newTxnManager(1, input)
+	output := txnManager.run()
+	select {
+	case input <- txnSmall:
+	case <-time.After(50 * time.Microsecond):
```
suzaku: Use `default`?

pkg/loader/load_test.go
lichunzhu: The 5th txn was picked from `input` but can't be added to the `txnManager.cacheChan`. Although it gets blocked at cond.Wait() but it's still picked out from `input`. By the way, we can't know a `txn`'s size when it is not picked out from `input`.

lichunzhu: The check `c.Assert(output, check.HasLen, 4)` can make sure it's not added.

pkg/loader/load_test.go
lichunzhu: This place `default` can be used.

pkg/loader/load_test.go
lichunzhu: I think wait some time is better. Since `input` has no buffer if `txnManager` is not exactly [waiting to get item from `input`](https://github.com/pingcap/tidb-binlog/pull/737/files#diff-b39261d071d61502be586d1b85e6dfb5R667) it can be passed. But in fact txnManager may still be busy popping items or doing some other things when we check here.

pkg/loader/load_test.go
lichunzhu: In `txnManager.run()` it will start a new goroutine and start watching input channel which will take some time. Using `default` may get fault.

pkg/loader/load_test.go
```diff
@@ -413,13 +413,13 @@ func (s *txnManagerSuite) TestRunTxnManager(c *check.C) {
 	case t := <-output:
 		txnManager.pop(t)
 		c.Assert(t, check.DeepEquals, txn)
-	case <-time.After(50 * time.Microsecond):
+	default:
 		c.Fatal("Fail to pick txn from txnManager")
 	}
```
suzaku:
```go
select {
case t := <-output:
case <-time.After...
}
```

This pattern appears a lot in this test, is it possible to extract this as a helper function to simplify the test?

lichunzhu:
@suzaku PTAL

pkg/loader/load_test.go
GregoryIan: it's more acurrate to use `default` here, because  `the 5th txn should get blocked at cond.Wait()` or `after <- input and before cond.Wait()`

GregoryIan:
Rest LGTM

july2993:

lichunzhu:

lichunzhu:

PR #743, opened 2019-09-05
Local

chenyulin4:
currently when we parse the kafka message of binlog, we can 
get the mysql data type of the data column from the field MysqlType. However, the detailed information of data type, like the length of string type or the precision of decimal type is missing. 
By the origin design, the detailed data types  of columns are supposed to be fetched from DDL messages, while in most cases , the DDL messages are not always in the topic due to the limitation of retention hours. 
Without the detailed data types, in the target database, we could not create the table with same definition, or add one column in the target table.

when putting binlog messages to kafka, the detailed data types of columns are added into the MysqlType field with the format like this: varchar(50)| decimal(3,2), etc.
By parsing the data type of the column ,the consumer could rebuild the column or the entire table with the same definition.
### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes


july2993:
pls create new issue instead of such pr

PR #744, opened 2019-09-05
tests: Rewrite updatePKUK to avoid outputing a lot of messages about generating PKs

suzaku:

updatePKUK output a lot of logs for retrying to generate unused PKs.

Record free PKs explicitly to avoid retrying.

### Check List 

Tests 

 - Integration test

Code changes

Side effects

Related changes

### DIFF
--- a/tests/dailytest/case.go
+++ b/tests/dailytest/case.go
@@ -351,34 +351,40 @@ func updatePKUK(db *sql.DB, opNum int) error {
 	maxKey := 20
 	mustExec(db, "create table pkuk(pk int primary key, uk int, v int, unique key uk(uk));")
 
-	var pks []int
+	pks := make(map[int]struct{})
+	freePks := make([]int, maxKey)
+	for i := 0; i < maxKey; i++ {
+		freePks[i] = i
+	}
+
+	nextPk := func() int {
+		return freePks[0]
+	}
 	addPK := func(pk int) {
-		pks = append(pks, pk)
+		pks[pk] = struct{}{}
+		freePks = freePks[1:]
 	}
 	removePK := func(pk int) {
-		var tmp []int
-		for _, v := range pks {
-			if v != pk {
-				tmp = append(tmp, v)
-			}
-		}
-		pks = tmp
+		delete(pks, pk)
+		freePks = append(freePks, pk)
 	}
-	hasPK := func(pk int) bool {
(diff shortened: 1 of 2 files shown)


tests/run.sh
```diff
@@ -161,7 +161,7 @@ run_case() {
 }
 
 # List the case names to run, eg. ("binlog" "kafka")
-do_cases=()
+do_cases=("binlog")
```
july2993: --?

suzaku: Fixed

tests/dailytest/case.go
```diff
@@ -351,34 +351,40 @@ func updatePKUK(db *sql.DB, opNum int) error {
 	maxKey := 20
 	mustExec(db, "create table pkuk(pk int primary key, uk int, v int, unique key uk(uk));")
 
-	var pks []int
+	pks := make(map[int]struct{})
+	freePks := make([]int, maxKey)
```
lichunzhu: How about `freePks := rand.Perm(maxKey)`?

suzaku: Cool, I didn't know about that, thanks.

lichunzhu:
LGTM

tests/dailytest/case.go
```diff
@@ -351,34 +351,37 @@ func updatePKUK(db *sql.DB, opNum int) error {
 	maxKey := 20
 	mustExec(db, "create table pkuk(pk int primary key, uk int, v int, unique key uk(uk));")
 
-	var pks []int
+	pks := make(map[int]struct{})
+	freePks := rand.Perm(maxKey)
+
+	nextPk := func() int {
+		return freePks[0]
```
july2993: https://godoc.org/math/rand#Shuffle before return first one or randomly pick one?

suzaku: @july2993 `rand.Perm` is enough? On deletion, the pks are deleted in random order and appended back to this slice.

july2993: there's no way you will do thinks like this:
update pkuk set pk = 1 where pk = 2;
update pkuk set pk = 2 where pk = 1;
update pkuk set pk = 1 where pk = 2;
update pkuk set pk = 2 where pk = 1;
...
continually to swap the two values, why not just simply make it more randomly?

PR #742, opened 2019-09-05
reparo/:Add worker-count, txn-batch config

lichunzhu:
### What problem does this PR solve? 
[TOOL-1547](https://internal.pingcap.net/jira/browse/TOOL-1547)
Add `worker count` and `txn batch` config for reparo. Users can decide these indecies themselves to change the sync speed of `reparo`.

### What is changed and how it works?
Expose `worker count` and `txn batch` to config.

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

Side effects

Related changes

 - Need to update the documentation
 - Need to be included in the release note

PR #746, opened 2019-09-09
reparo/:Add worker-count, txn-batch config (#742)

lichunzhu:
cherry-pick https://github.com/pingcap/tidb-binlog/pull/742
[TOOL-1547](https://internal.pingcap.net/jira/browse/TOOL-1547)
Add `worker count` and `txn batch` config for reparo. Users can decide these indecies themselves to change the sync speed of `reparo`.

Expose `worker count` and `txn batch` to config.

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

Side effects

Related changes

### DIFF
--- a/cmd/reparo/reparo.toml
+++ b/cmd/reparo/reparo.toml
@@ -19,6 +19,14 @@ log-level = "info"
 # for print, it just prints decoded value.
 dest-type = "mysql"
 
+# number of binlog events in a transaction batch
+txn-batch = 20
+
+# work count to execute binlogs
+# if the latency between reparo and downstream(mysql or tidb) are too high, you might want to increase this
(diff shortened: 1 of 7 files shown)


huynhphuong10284:
Just ask a question about "txn-batch" (default = 20) as below:
- If my system is big data, it has an issue that data will be delayed for a long time.
- If my system is small data, this value may be impacted to data delays.

The concern is "how to define a good value for any case in my system?"

lichunzhu:
@huynhphuong10284 It's not easy to "define a good value for any case in my system". For most of the time the default setting is enough. If you found that the latency is too high you can try increasing `worker-count`. If it doesn't work you can ask help from us.

PR #747, opened 2019-09-09
optimize drainer syncer quit

lichunzhu:
In `syncer.run` if we quit directly channel `syncer.closed` will not be closed and drainer will get blocked while closing.

Change `return` to `break ForLoop`

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

Side effects

Related changes


lichunzhu:
> if need this, also need chang L380 & L387?
> 
> what's the root problem you encounter?

I was checking Shaowen Yin's drainer log yesterday and find that drainer can't successfully quit.
What do you mean *chang L380 & L387*?

PR #758, opened 2019-09-30
pump.Storage: fix possible panic if start ts is greater than max commit ts when pull binlog 

WangXiangUSTC:
in pull binlog, if start ts is greater than max commit ts, pump may panic because goleveldb's bug https://github.com/syndtr/goleveldb/issues/224 

add a check in storage
and update some log by the way

### Check List 

Tests 

 - Manual test

Related changes

### DIFF
--- a/pump/storage/sorter.go
+++ b/pump/storage/sorter.go
@@ -15,6 +15,7 @@ package storage
 
 import (
 	"container/list"
+	"fmt"
 	"math/rand"
 	"sync"
 	"sync/atomic"
@@ -90,6 +91,11 @@ type sortItem struct {
 	start  int64
 	commit int64
 	tp     pb.BinlogType
+}
+
+// String implements fmt.Stringer
+func (s *sortItem) String() string {
+	return fmt.Sprintf("{start: %d, commit: %d, tp: %v}", s.start, s.commit, s.tp)
 }
 
 type sorter struct {
--- a/pump/storage/storage.go
+++ b/pump/storage/storage.go
@@ -1110,6 +1110,10 @@ func (a *Append) PullCommitBinlog(ctx context.Context, last int64) <-chan []byte
 		Limit: encodeTSKey(math.MaxInt64),
 	}
 
+	pLog := pkgutil.NewLog()
+	labelWrongRange := "wrong range"
+	pLog.Add(labelWrongRange, 10*time.Second)
+
 	go func() {
 		defer close(values)
 
@@ -1117,8 +1121,11 @@ func (a *Append) PullCommitBinlog(ctx context.Context, last int64) <-chan []byte
 			startTS := last + 1
 			limitTS := atomic.LoadInt64(&a.maxCommitTS) + 1
 			if startTS > limitTS {
-				log.Warn("last ts is greater than pump's max commit ts", zap.Int64("last ts", startTS-1), zap.Int64("max commit ts", limitTS-1))
-				time.Sleep(time.Second * 10)
+				// if range's start is greater than limit, may cause panic, see https://github.com/syndtr/goleveldb/issues/224 for detail.
+				pLog.Print(labelWrongRange, func() {
+					log.Warn("last ts is greater than pump's max commit ts", zap.Int64("last ts", startTS-1), zap.Int64("max commit ts", limitTS-1))
+				})
+				time.Sleep(time.Second)
 				continue
 			}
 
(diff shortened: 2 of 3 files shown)


pump/storage/sorter.go
```diff
@@ -92,6 +93,11 @@ type sortItem struct {
 	tp     pb.BinlogType
 }
 
+// String implements fmt.Stringer
+func (s *sortItem) String() string {
+	return fmt.Sprintf("{start: %d, commit: %d, tp: %v}", s.start, s.commit, s.tp)
```
GregoryIan:
```suggestion
	return fmt.Sprintf("{start: %d, commit: %d, tp: %s}", s.start, s.commit, s.tp)
```

pump/storage/storage.go
```diff
@@ -1114,9 +1115,15 @@ func (a *Append) PullCommitBinlog(ctx context.Context, last int64) <-chan []byte
 
 		for {
 			startTS := last + 1
+			limitTS := atomic.LoadInt64(&a.maxCommitTS) + 1
+			if startTS > limitTS {
+				log.Warn("last ts is greater than pump's max commit ts", zap.Int64("last ts", startTS-1), zap.Int64("max commit ts", limitTS-1))
+				time.Sleep(time.Second * 10)
```
GregoryIan: 10 second is too long, how about return directly

GregoryIan: if so, you also need to check logic of drainer

pump/storage/vlog.go
```diff
@@ -120,6 +120,20 @@ func (r *request) String() string {
...
 }
 
+type batchRequest []*request
+
+// String implements fmt.Stringer
+func (b *batchRequest) String() string {
+	s := new(strings.Builder)
+	s.WriteString("{")
+	for _, r := range []*request(*b) {
+		s.WriteString(r.String())
```
GregoryIan: break line or `comma`?

GregoryIan:
Rest LGTM

pump/storage/storage.go
WangXiangUSTC: can't return here, otherwise, drainer will not get binlog forever and need restart

pump/storage/vlog.go
WangXiangUSTC: add a comma

pump/storage/storage.go
WangXiangUSTC: and this need update a lot of code,  I think just wait is ok

GregoryIan: 10 second is too long

pump/storage/storage.go
```diff
@@ -1114,9 +1115,15 @@ func (a *Append) PullCommitBinlog(ctx context.Context, last int64) <-chan []byte
 
 		for {
 			startTS := last + 1
+			limitTS := atomic.LoadInt64(&a.maxCommitTS) + 1
+			if startTS > limitTS {
```
july2993: just set `limitTS  to be startTS + 1 if  startTS >= limitTS` to by pass the bug
don't log warn log and sleep 10 second

WangXiangUSTC: will it cause some bug if query some binlog is greater than the max commit ts?

pump/storage/storage.go
WangXiangUSTC: update to 1 second

pump/storage/storage.go
WangXiangUSTC: add a comment

GregoryIan:
LGTM

pump/storage/storage.go
july2993: correct:
set  limitTS to be startTS  if startTS > limitTS if the bug about goleveldb will only trigger when
`start > limit`

[startTS,   limitTS( = startTS)     will get no any binlog

pump/storage/storage.go
```diff
@@ -1109,14 +1110,27 @@ func (a *Append) PullCommitBinlog(ctx context.Context, last int64) <-chan []byte
...
 	go func() {
 		defer close(values)
 
 		for {
 			startTS := last + 1
+			limitTS := atomic.LoadInt64(&a.maxCommitTS) + 1
+			if startTS > limitTS {
+				// if range's start is greater than limit, may cause panic, see https://github.com/syndtr/goleveldb/issues/224 for detail.
+				pLog.Print(labelWrongRange, func() {
+					log.Warn("last ts is greater than pump's max commit ts", zap.Int64("last ts", startTS-1), zap.Int64("max commit ts", limitTS-1))
```
july2993: This is actually a normal case, so don't add warn log.

like the MaxCommitTS = 10 in pump, checkpointTS = 10 for drainer,
so check for if we having newer binlogs by [checkpointTS + 1, MaxCommitTS + 1)

WangXiangUSTC: in normal case startTS is equal to limitTS, but will not greater than it?

july2993: look yes, but why will trigger this?

july2993:

PR #773, opened 2019-10-17
pump/storge: Measure time taken to write to disk, to rotate and fsync

suzaku:

We only measurer the time taken to write a batch of binlogs at the top level.
But we have no other measurementt to help us figure out which step is taking up most of the time.

Add 3 metrics:

1. Time taken to write a buf
1. Time taken to rotate
1. Time taken to fsync

### Check List 

Tests 

 - Unit test

Code changes

Side effects

Related changes

### DIFF
--- a/pump/storage/log.go
+++ b/pump/storage/log.go
@@ -22,6 +22,7 @@ import (
 	"os"
 	"sync"
 	"sync/atomic"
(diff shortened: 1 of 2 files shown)


GregoryIan:
Good Job! LGTM

PR #778, opened 2019-10-22
kafka.go: Refine log

july2993:
refine log.

log fatal instead of panic(err), so user can know what really happen
instead of like drainer restart by some unknown reason before check
stderr.

### Check List 

Tests


july2993:
> LGTM.
> 
> It is still expected that Kafka error should kill Drainer, right?

yes

pkg/util/log.go
```diff
@@ -65,8 +65,7 @@ func InitLogger(level string, file string) error {
 	cfg := &log.Config{
 		Level: level,
 		File: log.FileLogConfig{
-			Filename:  file,
-			LogRotate: true,
```
july2993: no any affect field in older version and is removed now.

PR #799, opened 2019-11-07
Fix panic of empty value in filter config

july2993:
Fix panic of empty value in filter config
Verify config about filters

Fix panic of empty value in filter config
Verify config about filters

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

Side effects

Related changes

### DIFF
--- a/pkg/filter/filter.go
+++ b/pkg/filter/filter.go
@@ -48,10 +48,10 @@ func NewFilter(ignoreDBs []string, ignoreTables []TableName, doDBs []string, doT
 func (s *Filter) addOneRegex(originStr string) {
 	if _, ok := s.reMap[originStr]; !ok {
 		var re *regexp.Regexp
-		if originStr[0] != '~' {
-			re = regexp.MustCompile(fmt.Sprintf("(?i)^%s$", originStr))
-		} else {
+		if len(originStr) > 0 && originStr[0] == '~' {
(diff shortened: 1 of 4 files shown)


pkg/filter/filter.go
```diff
@@ -48,10 +48,10 @@ func NewFilter(ignoreDBs []string, ignoreTables []TableName, doDBs []string, doT
 func (s *Filter) addOneRegex(originStr string) {
 	if _, ok := s.reMap[originStr]; !ok {
 		var re *regexp.Regexp
-		if originStr[0] != '~' {
-			re = regexp.MustCompile(fmt.Sprintf("(?i)^%s$", originStr))
-		} else {
+		if len(originStr) > 0 && originStr[0] == '~' {
 			re = regexp.MustCompile(fmt.Sprintf("(?i)%s", originStr[1:]))
+		} else { // must match completely
+			re = regexp.MustCompile(fmt.Sprintf("(?i)^%s$", originStr))
```
suzaku: If `originStr` is empty, the regex become `^$`?

july2993: yes, the else case means you must match totally.
but we check the config for empty schema and table at startup time, so it will not run into this case now actually, because it's meaningless for an empty schema and table, may cause by misconfiguring.

PR #802, opened 2019-11-11
Fix panic of empty value in filter config (#799)

july2993:
Fix panic of empty value in filter config
Verify config about filters with empty schema or table name.

cherry-pick #799 

Fix panic of empty value in filter config
Verify config about filters with empty schema or table name.

### Check List 

Tests 

 - Unit test

Code changes

Side effects

Related changes


GregoryIan:
LGTM

PR #803, opened 2019-11-11
drainer: bugfix, handle "missing column" when a column is getting dropped

suzaku:

DML binlogs may be wrong if there's a concurrent `drop column` DDL running.
The column being dropped may be missing before the corresponding DDL binlog reached drainer, because TiDB stops sending the dropped column starting from the DeleteOnly state.

1. Add an integration test to expose the problem.
2. TiDB would send an extra DDL binlog when the job reaches DeleteOnly state pingcap/tidb#13536
3. Drainer use this extra binlog to decide that a table is dropping column
4. If a column is missing while the table is dropping column, we fill the column with default values.

### Check List 

Tests 

 - Integration test

Code changes

Side effects

Related changes


GregoryIan:
LGTM now

ericsyh:
@suzaku pls fix the conflicts.

WangXiangUSTC:
LGTM

WangXiangUSTC:
need to add release note, and update upgrade document(may not compatible with old tidb?)

suzaku:
> need to add release note, and update upgrade document(may not compatible with old tidb?)

1. Where can I add a release note?
1. It's compatible with older TiDB, with no extra binlog it works the same way as before (with the bug). But older Drainers is not compatible with newer TiDBs because they can't handle the extra binlog correctly.

PR #806, opened 2019-11-13
fix deadlock while drainer.Syncer panic

GregoryIan:
we start `Syncer` goroutine in `drainer`
```
s.tg.GoNoPanic("syncer", func() {
	defer func() { go s.Close() }()
	if err := s.syncer.Start(); err != nil {
		log.Error("syncer exited abnormal", zap.Error(err))
	}
})
```

if `s.syncer` panic (like kafka producer panic), `s.syncer.Start()` would exit directly and  the logics will happen 
* drainer calls `s.syncer.Close()` in `s.Close()`
* `s.syncer.Close()` would wait `<- s.syncer.closed` take effetcs,
* but `s.syncer.closed` only can be closed in `s.syncer.run()`

add `defer close(s.syncer.closed)` in `s.syncer.run()`

### Check List

### DIFF
--- a/drainer/syncer.go
+++ b/drainer/syncer.go
@@ -258,6 +258,8 @@ func (s *Syncer) savePoint(ts, slaveTS int64) {
 }
 
 func (s *Syncer) run() error {
+	defer close(s.closed)
+
 	wait := make(chan struct{})
 
 	fakeBinlogCh := make(chan *pb.Binlog, 1024)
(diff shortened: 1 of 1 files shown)


GregoryIan:
I want to add an integration test, but integration test framework of `TiDB-Binlog` doesn't have many utils function, like `check liveness of drainer process`, so I give up

suzaku:
> I want to add an integration test, but integration test framework of `TiDB-Binlog` doesn't have many utils function, like `check liveness of drainer process` and `failpoint`, so I give up, it's too weak 😩

I've created an issue about adding a related test on Jira. LGTM for now.

PR #809, opened 2019-11-14
loader: add a log when table not found

leoppro:
add a log when table not found

when table not found in `getTableInfo` function, we can't know which table was unfound.

### Check List 

Tests 

 - Unit test
 - Integration tes

### DIFF
--- a/pkg/loader/util.go
+++ b/pkg/loader/util.go
@@ -21,6 +21,8 @@ import (
 	"strings"
 
 	"github.com/pingcap/errors"
+	"github.com/pingcap/log"
+	"go.uber.org/zap"
 )
 
 var (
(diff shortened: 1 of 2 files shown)


pkg/loader/util.go
```diff
@@ -58,6 +60,7 @@ func getTableInfo(db *gosql.DB, schema string, table string) (info *tableInfo, e
 
 	if info.columns, err = getColsOfTbl(db, schema, table); err != nil {
 		if err == ErrTableNotExist {
+			log.Warn("table not exist", zap.String("schema", schema), zap.String("table", table))
 			return nil, err
```
july2993: use errors.Annotatef  instead of a the warn log

PR #812, opened 2019-11-19
drainer: remove deprecated syncer flash

suzaku:

1. The flash syncer is deprecated and is seldom used.
1. It takes a lot of effort to take care of the flash syncer when refactoring components of drainer

Remove the obsolete syncer.

### Check List 

Tests 

 - Unit test

Code changes

    A dest db type is removed.

Side effects

    Users of new version of binlog including this change will not be able to use flash as a dest db.

Related changes


GregoryIan:
need to update document, and maybe also ansible

GregoryIan:
LGTM

suzaku:
@july2993 PTAL

PR #813, opened 2019-11-19
add timeout for gs.GracefulStop

GregoryIan:
sometime pump would block at `gs.GracefulStop` when it exits, we still don't know the specific reason, but we don't need to wait for GRPC service exited, it's safe.

add a timeout mechanism to give up waiting for the goroutine to exit
### Check List 

Tests 

 - Unit test


pkg/util/util.go
```diff
@@ -280,3 +280,22 @@ func AdjustDuration(v *time.Duration, defValue time.Duration) {
...
 	}
 }
+
+// GoAndAbortGoroutine creates a goroutine to run fn, and then gives up waiting for the goroutine to exit When it timeouts
+func GoAndAbortGoroutine(name string, fn func(), timeout time.Duration) {
+	fName := zap.String("name", name)
+	exited := make(chan struct{})
+	go func() {
+		defer func() {
+			log.Info("goroutine exit bt itself (with GoAndAbortGoroutine help)", fName)
```
suzaku: What does `bt` mean?

pkg/util/util.go
```diff
@@ -280,3 +280,22 @@ func AdjustDuration(v *time.Duration, defValue time.Duration) {
...
+			close(exited)
+		}()
+		fn()
+	}()
+
+	select {
+	case <-time.After((timeout)):
```
suzaku:
```suggestion
	case <-time.After(timeout):
```

pkg/util/util.go
```diff
@@ -280,3 +280,22 @@ func AdjustDuration(v *time.Duration, defValue time.Duration) {
 		*v = defValue
 	}
 }
+
+// GoAndAbortGoroutine creates a goroutine to run fn, and then gives up waiting for the goroutine to exit When it timeouts
+func GoAndAbortGoroutine(name string, fn func(), timeout time.Duration) {
```
suzaku: Maybe `WaitUntilTimeout` is a better name?

PR #815, opened 2019-11-20
refine notifying drainer

IANTHEREAL:
New pump will notify drainer to pull the binlog in time before providing service.
But sometimes the request fails because the drainer meet some trouble, like fail to request PD to fetch the status of all pump instance

behavior of functions:
* merger.prepareChangingSource: set source changing
* merger.AddSource: add source -> merger.SetSourceChanged
* merger.RemoveSource； remove source -> merger.SetSourceChanged
* merger.SetSourceChanged: set source changed -> set !source changing

temporal logic of binlog:
pump:
* set source changing(collctor) -> new pump writes binlog

collector:
* set source changing -> add source -> set source changed -> set !source changing

merger:
* set !source changed -> find binlog from all sources -> read !source changing -> read !source changed -> publish binlog

the only wrong temporal logics meets two conditions at the same time
find binlog(merger) -> add source(collector) -> read !source changing(merger) -> read !source changed(merger) -> read !source changed(merger) -> publish binlog(merger)
* new pump writes binlog -> find binlog(merger)

Let's analyze the possible situations:

if set source changed(collector) -> read !source changed(merger), it means
* set source changed(collector) -> set !source changed(merger)
* add source(collector) -> set source changed(collector) &&
* set !source changed(merger) -> find binlog from all sources(merger)
=> add source(client) -> find binlog(merger), it's safe!!!

and

if read !source changed(merger) -> set source changed(collector), there are two conditions:
1. if read !source changing(merger) -> set source changing(collector), it means
	* find binlog(merger) -> read !source changing(merger) &&
	* set source changing(collector) -> write new binlog
	=> find binlog(merger) -> write new binlog, it's safe!!!

2. if set source changing(collector) -> read !source changing(merger), it means
	* set source changed(collector) -> set source changing(collector) -> read !source changing(merger) &&
	* read !source changing(merger) -> read !source changed(merger)
	=> set source changed(collector) -> read !source changed(merger), Contradictory with assumption


drainer/collector.go
```diff
@@ -344,11 +344,24 @@ func (c *Collector) keepUpdatingStatus(ctx context.Context, fUpdate func(context
...
+			// stop merge and return directly
+			// temporal logic:
+			// * merge is paused -> new pump write binlog
+			// * fetch all sources in merge -> find smallest commit ts binlog -> check merge whether paused -> publish binlog
+			// so it's safe
+			// TODO: maybe we can remove merge.SourceChanged
+			c.merger.Stop()
 			nr.wg.Done()
+			if err := fUpdate(ctx); err != nil {
+				log.Error("Failed to update collector status (notifying drainer progress)", zap.Error(err))
```
WangXiangUSTC: maybe we can add a failpoint test here later

GregoryIan: tidb binlog doesn't have failpoint

suzaku: We don't need to use `failpoint` here because `fUpdate` is a parameter you can pass. To test error handling we can just pass in a function that returns error as we like.

GregoryIan: there're not any error handling here now, we need to test the temporal logic, it needs failpoint

GregoryIan:
@july2993 @suzaku PTAL, this PR is too old

IANTHEREAL:
I close this pr firstly and  will open it while we need it

PR #823, opened 2019-11-22
avoid oom when there is lag and binlog size is big

july2993:
try best to avoid OOM when msg is big for drainer&arbiter

- stall to produce new me msg into producer when the total size of messages waiting to be acked reaching a limit value.

this can help avoid OOM, but no performance back with a reasonable value.

- remove some buffer in `Reader`

note we will read msg form `Reader` ASAP, and handle how many events to buffer in `Loader`
so we can just set one for the chan buffer size in the Reader.
this help avoids arbiter oom in some case.

block by https://github.com/pingcap/tidb-tools/pull/291

### Check List 

Tests 
testing with binlog size = 40M
before this pr, drainer will reach more than 130GiB, after this pr, only a few GiB.
for the arbiter, it will slow to consume so no buffer cause oom in my env, maybe need more investigate. But no performance back when the size of binlog is small.


drainer/sync/kafka.go
```diff
@@ -29,6 +29,7 @@ import (
 )
 
 var maxWaitTimeToSendMSG = time.Second * 30
+var stallWriteSize = 90 * 1024 * 1024
```
suzaku: How was this size decided?

july2993: internaly of sarama.

there is one in-memory produceSet buffer and maybe one inflight request.

with the current config of the flush strategy only setting `Flush.MaxMessage`(kafka-max-messages of drainer's config)
it will roll over the in-memory produceSet and send it ASAP once there's no inflight request.
and continually consuming the messages and push into the in-memory produceSet(but once reach some limit, message number larger than `Flush.MaxMessge` or total size large than `sarama.MaxRequest` size here), it will wait for space, means waiting for response of the inflight request and send this in-memory request in one request.


Here we just need a large enough value for a batch that should be ok.

besides
when sending a request big request size may cause this:
```
[2019-11-25 00:15:23,693] WARN [SocketServer brokerId=0] Unexpected error from /0:0:0:0:0:0:0:1; closing connection (org.apache.kafka.common.network.Selector)
org.apache.kafka.common.network.InvalidReceiveException: Invalid receive (size = 411517920 larger than 104857600)
        at org.apache.kafka.common.network.NetworkReceive.readFrom(NetworkReceive.java:104)
        at org.apache.kafka.common.network.KafkaChannel.receive(KafkaChannel.java:424)
        at org.apache.kafka.common.network.KafkaChannel.read(KafkaChannel.java:385)
        at org.apache.kafka.common.network.Selector.attemptRead(Selector.java:640)
        at org.apache.kafka.common.network.Selector.pollSelectionKeys(Selector.java:561)
        at org.apache.kafka.common.network.Selector.poll(Selector.java:472)
        at kafka.network.Processor.poll(SocketServer.scala:830)
        at kafka.network.Processor.run(SocketServer.scala:730)
        at java.lang.Thread.run(Thread.java:748)
```
the relate config of broker is  `socket.request.max.bytes` which is default 100M.

july2993: note if we setting `Flush.MaxMessage` to be a small value to reduce memory when the message size is big that should be ok.
But when the message size is small it will not batch too many messages in one request, but can limit the total size of messages push into the producer  to control the maximum size of one
request for two different workloads. Of course, if one message's size bigger than `stallWriteSize`, then the request size containing only one message will bigger than `stallWriteSize`.

GregoryIan:
LGTM

PR #830, opened 2019-11-25
pump/storge: Measure time taken to write to disk, to rotate and fsync

suzaku:

Cherry-pick pingcap/tidb-binlog#773


GregoryIan:
LGTM

PR #774, opened 2019-10-17
binlogctl: Only show offline nodes when explicitly told so

suzaku:
### What problem does this PR solve? 

Some users complain that pump/drainer nodes that have been turn off still appears in the list.

### What is changed and how it works?

Hide `offline` nodes by default, only show them when `-show-offline-nodes` is set.

### Check List 

Tests 

 - Manual test
    I set a drainer to `offline` and then check with `binlogctl drainers`.

Code changes

Side effects

Related changes

 - Need to update the documentation

PR #832, opened 2019-11-25
binlogctl: Only show offline nodes when explicitly told so (#774)

suzaku:

Cherry-pick #774


GregoryIan:
LGTM

suzaku:
Integration tests should pass after merging #835 .

PR #834, opened 2019-11-25
avoid oom when there is lag and binlog size is big (#823)

july2993:
cherry-pick #823

### Check List 

Tests


GregoryIan:
LGTM

PR #837, opened 2019-12-03
arbiter: discard the repeated kafka message

leoppro:
we can‘t ensure that the message in kafka is unique, so we need discard the repeated kafka message

### Check List 

Tests 

 - Unit test


arbiter/server.go
```diff
@@ -287,8 +287,15 @@ func (s *Server) loadStatus() (int, error) {
 func syncBinlogs(ctx context.Context, source <-chan *reader.Message, ld loader.Loader) (err error) {
 	dest := ld.Input()
 	defer ld.Close()
+	var receivedTs int64
 	for msg := range source {
 		log.Debug("recv msg from kafka reader", zap.Int64("ts", msg.Binlog.CommitTs), zap.Int64("offset", msg.Offset))
+
+		if msg.Binlog.CommitTs <= receivedTs {
```
suzaku: Add a log message for the ignored commit ts?

leoppro: updated

PR #840, opened 2019-12-05
*: Add missing error checks and enable errcheck in static-check

suzaku:

The errcheck analysis is disabled because there are still many places where the returned error is unchecked.

Add missing error checks and enable errcheck.

### Check List 

Tests 

 - Unit test

### DIFF
--- a/pump/storage/log.go
+++ b/pump/storage/log.go
@@ -369,7 +369,11 @@ func seekToNextRecord(reader *bufio.Reader) (bytes int, err error) {
 			return
 		}
 
-		reader.Discard(1) //nolint: errcheck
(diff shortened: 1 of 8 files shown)


pump/storage/log.go
```diff
@@ -369,7 +369,7 @@ func seekToNextRecord(reader *bufio.Reader) (bytes int, err error) {
 			return
 		}
 
-		reader.Discard(1)
+		reader.Discard(1) //nolint: errcheck
```
july2993: Add comment for why it will never happen.
and just panic or `var _ = reader.Discard(1)` ?

suzaku: Fixed.

Issue #839, opened 2019-12-05
Summer time increased by 1 hour through pump & drainer


you06:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

1. What did you do?

* road
```
tidb -> pump -> drainer -> tidb
```

* Create table and insert some summer time records in source database.

```
CREATE TABLE `t` (`d` timestamp NOT NULL);
INSERT INTO t(d) VALUES('1988-05-28 05:58:57'), ('2011-11-11 11:11:11');
```

* Select from source database

```
mysql> SELECT d, UNIX_TIMESTAMP(d) FROM t;
+---------------------+-------------------+
| d                   | UNIX_TIMESTAMP(d) |
+---------------------+-------------------+
| 1988-05-28 05:58:57 |         580773537 |
| 2011-11-11 11:11:11 |        1320981071 |
+---------------------+-------------------+
2 rows in set (0.00 sec)
```

* Select from target database

```
mysql> SELECT d, UNIX_TIMESTAMP(d) FROM t;
+---------------------+-------------------+
| d                   | UNIX_TIMESTAMP(d) |
+---------------------+-------------------+
| 2011-11-11 11:11:11 |        1320981071 |
| 1988-05-28 06:58:57 |         580777137 |
+---------------------+-------------------+
2 rows in set (0.00 sec)
```

2. What did you expect to see?

`1988-05-28 05:58:57` instead of `1988-05-28 06:58:57` in target db.

3. What did you see instead?

`1988-05-28 06:58:57` in target db.

4. Please provide the relate downstream type and version of drainer.
(run `drainer -V` in terminal to get drainer's version)

```
# ./drainer -V
Release Version: v3.0.6
Git Commit Hash: 948d8f32643321a038a79c91effcf55246d41e15
Build TS: 2019-12-04 09:30:14
Go Version: go1.13
Go OS/Arch: linux/amd64
```

leoppro:
fixed in #841

PR #841, opened 2019-12-05
fix time zone problem, unify the time zone with downstream

leoppro:
Issue：#839
When the system time zone of drainer is different from the downstream database, the timestamp value will be parsed mistakenly.

In drainer/arbiter/reparo, we parse time using UTC time zone, and set the session time zone of downstream database UTC.
And add a field in file and kafka protobuf to marks if the timestamp is in UTC.

### Check List 

Tests 

 - Unit test
 - Manual test


drainer/translator/kafka.go
```diff
@@ -126,7 +126,7 @@ func deleteRowToRow(tableInfo *model.TableInfo, raw []byte) (row *obinlog.Row, e
 	columns := tableInfo.Columns
 
 	colsTypeMap := util.ToColumnTypeMap(tableInfo.Columns)
-	columnValues, err := tablecodec.DecodeRow(raw, colsTypeMap, time.Local)
+	columnValues, err := tablecodec.DecodeRow(raw, colsTypeMap, time.UTC)
```
suzaku: This change is not backward compatible, right?

leoppro: yes

suzaku: OK, we should make sure to document the behavior change since the version including this change.

july2993: don't suggest make such a breaking change for down-type = kafka & file
maybe add a field to indicate the time zone of time.Local ?

GregoryIan: how to deal with this problem now @leoppro?

leoppro: I will add a flag for `down-type = kafka & file` to avoid breaking compatibility

leoppro: fixed,PTAL again @july2993

leoppro:
PTAL again @july2993

leoppro:
I will add test case for timestamp value

GregoryIan:
please make sure this PR alive, otherwise I will close it

Issue #768, opened 2019-10-16
TiDB Binlog high delay


billhcmus:
Hi,
I'm testing TiDB Binlog in real-server but got high delay metrics reported. Detail:

- Binlog Reach Duration By Pump metrics: greater than 1 second

<img width="646" alt="Screen Shot 2019-10-16 at 15 41 47" src="https://user-images.githubusercontent.com/16277024/66903954-8231b100-f02d-11e9-8427-c6c910ab1d1f.png">

- Transaction execution (sql query Time) greater than 1 second

<img width="650" alt="Screen Shot 2019-10-16 at 15 42 39" src="https://user-images.githubusercontent.com/16277024/66904159-d046b480-f02d-11e9-80d8-a53d1f4cc40f.png">

- Then the Checkpoint delay (p99) is 8.7 min and avg is 2.0 min

<img width="654" alt="Screen Shot 2019-10-16 at 15 44 05" src="https://user-images.githubusercontent.com/16277024/66904215-e6ed0b80-f02d-11e9-9e9c-02b8a18dffd5.png">

My Drainer configuration is:

```
....
txn-batch = 1000
worker-count = 48
flag: --cache-binlog-count=6553600
```
- So after 1 second Binlog reach to Drainer, but drainer need > 7s to execute this transaction? Is it right?

Hope your help.

july2993:
seems there's lag at that time near 15:10, may meet high throughput at that time at up stream or run a slow DDL? you can check `Drainer Event` and `ddl` at drainer dashboard for the speed it handled DDL and DML events at that time.


- suggest use a lower txn-batch (<= 100)
- don't config the `cache-binlog-count`, just use the default (with too high value may causes OOM when there's lag), it should be ok to just use the default value after #737  or v3.04

billhcmus:
Thank you so much.

PR #843, opened 2019-12-06
drainer: Reduce Binlog sync delay

billhcmus:

Reference this issue: https://github.com/pingcap/tidb-binlog/issues/768

* When we benchmark TiDB Binlog with:
- Upstream: 1 TiDB, 1 PD, 1 TiKV
- TiDB Binlog: 1 Pump, 1 Drainer
- Downstream: 1 TiDB, 1 PD, 1 TiKV
- Hardware:  Intel(R) Xeon(R) CPU E5-2620 0 @ 2.00GHz, 32GB RAM

* The delay recorded bellow:

image

- The binlog-sync-delay calculate when one binlog success apply to downstream, recorded in function: `handleSuccess`

- We benchmark with sysbench, with insert and update operation
- Upstream throughput show bellow:

```
[ 10s ] thds: 24 tps: 4212.96 qps: 4212.96 (r/w/o: 0.00/4212.96/0.00) lat (ms,99%): 15.00 err/s: 0.00 reconn/s: 0.00
...
[ 1200s ] thds: 24 tps: 4134.75 qps: 4134.75 (r/w/o: 0.00/4134.75/0.00) lat (ms,99%): 16.12 err/s: 0.00 reconn/s: 0.00
SQL statistics:
    queries performed:
        read:                            0
        write:                           5149001
        other:                           0
        total:                           5149001
    transactions:                        5149001 (4290.78 per sec.)
    queries:                             5149001 (4290.78 per sec.)
    ignored errors:                      0      (0.00 per sec.)
    reconnects:                          0      (0.00 per sec.)

General statistics:
    total time:                          1200.0097s
    total number of events:              5149001

Latency (ms):
         min:                                    2.00
         avg:                                    5.59
         max:                                  467.56
         99th percentile:                       12.98
         sum:                             28764457.03

Threads fairness:
    events (avg/stddev):           214541.7083/531.97
    execution time (avg/stddev):   1198.5190/0.10
```

* The problem is TiDB Binlog uses a lot of blocking queue (channel) in Syncer, loader, and translator,
after debug I have been found this problem. So I decided to optimize your version by replacing go channel and use LMAX Disruptor instead.
* The result after replacing:
image

* Currently I just replace go channel by disruptor and the txnManager has been removed because it's an unbuffed channel and current I have no idea with it (hope your custom). Some unit test relates txnManager has been removed.
* The Ringbuffer size I set is 2^16 (this not satisfied with big Txn, we can add to config later)

### Check List

Tests
 - Unit test

Code changes

Related changes

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -105,7 +105,7 @@ endif
 
 check-static: tools/bin/golangci-lint
 	$(GO) mod vendor
-	tools/bin/golangci-lint --disable errcheck run $$($(PACKAGE_DIRECTORIES))
+	tools/bin/golangci-lint run $$($(PACKAGE_DIRECTORIES))
 
 clean:
 	go clean -i ./...
--- a/drainer/server.go
+++ b/drainer/server.go
@@ -335,13 +335,19 @@ func (s *Server) ApplyAction(w http.ResponseWriter, r *http.Request) {
 	log.Info("receive apply action request", zap.String("nodeID", nodeID), zap.String("action", action))
 
 	if nodeID != s.ID {
-		rd.JSON(w, http.StatusOK, util.ErrResponsef("invalid nodeID %s, this pump's nodeID is %s", nodeID, s.ID))
+		err := rd.JSON(w, http.StatusOK, util.ErrResponsef("invalid nodeID %s, this pump's nodeID is %s", nodeID, s.ID))
+		if err != nil {
+			log.Error("Failed to render JSON response", zap.Error(err))
+		}
 		return
 	}
 
 	s.statusMu.RLock()
 	if s.status.State != node.Online {
-		rd.JSON(w, http.StatusOK, util.ErrResponsef("this pump's state is %s, apply %s failed!", s.status.State, action))
+		err := rd.JSON(w, http.StatusOK, util.ErrResponsef("this pump's state is %s, apply %s failed!", s.status.State, action))
+		if err != nil {
+			log.Error("Failed to render JSON response", zap.Error(err))
+		}
 		s.statusMu.RUnlock()
 		return
 	}
@@ -355,13 +361,19 @@ func (s *Server) ApplyAction(w http.ResponseWriter, r *http.Request) {
 		s.status.State = node.Closing
 	default:
 		s.statusMu.Unlock()
-		rd.JSON(w, http.StatusOK, util.ErrResponsef("invalid action %s", action))
+		err := rd.JSON(w, http.StatusOK, util.ErrResponsef("invalid action %s", action))
+		if err != nil {
+			log.Error("Failed to render JSON response", zap.Error(err))
+		}
 		return
 	}
 	s.statusMu.Unlock()
 
 	go s.Close()
-	rd.JSON(w, http.StatusOK, util.SuccessResponse(fmt.Sprintf("apply action %s success!", action), nil))
+	err := rd.JSON(w, http.StatusOK, util.SuccessResponse(fmt.Sprintf("apply action %s success!", action), nil))
+	if err != nil {
+		log.Error("Failed to render JSON response", zap.Error(err))
+	}
 }
 
 // GetLatestTS returns the last binlog's commit ts which synced to downstream.
@@ -370,7 +382,10 @@ func (s *Server) GetLatestTS(w http.ResponseWriter, r *http.Request) {
 		IndentJSON: true,
 	})
 	ts := s.syncer.GetLatestCommitTS()
-	rd.JSON(w, http.StatusOK, util.SuccessResponse("get drainer's latest ts success!", map[string]int64{"ts": ts}))
+	err := rd.JSON(w, http.StatusOK, util.SuccessResponse("get drainer's latest ts success!", map[string]int64{"ts": ts}))
+	if err != nil {
+		log.Error("Failed to render JSON response", zap.Error(err))
+	}
 }
 
 // commitStatus commit the node's last status to pd when close the server.
--- a/drainer/server_test.go
+++ b/drainer/server_test.go
@@ -346,8 +346,9 @@ func (s *newServerSuite) TestInvalidDestDBType(c *C) {
 	cfg.DataDir = path.Join(c.MkDir(), "drainer")
 	cfg.ListenAddr = "http://" + cfg.ListenAddr
 	cfg.SyncerCfg.DestDBType = "nothing"
-	cfg.adjustConfig()
-	_, err := NewServer(cfg)
+	err := cfg.adjustConfig()
+	c.Assert(err, IsNil)
+	_, err = NewServer(cfg)
 	c.Assert(err, ErrorMatches, ".*unknown DestDBType.*")
 	c.Assert(cfg.SyncerCfg.To.ClusterID, Equals, uint64(8012))
 }
--- a/drainer/status.go
+++ b/drainer/status.go
@@ -16,6 +16,9 @@ package drainer
 import (
 	"encoding/json"
 	"net/http"
+
+	"github.com/pingcap/log"
+	"go.uber.org/zap"
 )
 
 // HTTPStatus exposes current status of the collector via HTTP
@@ -28,5 +31,7 @@ type HTTPStatus struct {
 
 // Status implements http.ServeHTTP interface
 func (s *HTTPStatus) Status(w http.ResponseWriter, r *http.Request) {
-	json.NewEncoder(w).Encode(s)
+	if err := json.NewEncoder(w).Encode(s); err != nil {
+		log.Error("Failed to encode status", zap.Error(err), zap.Any("status", *s))
+	}
 }
--- a/pump/node_test.go
+++ b/pump/node_test.go
@@ -22,6 +22,9 @@ import (
 	"strings"
 	"time"
 
+	"github.com/pingcap/log"
+	"go.uber.org/zap"
+
 	. "github.com/pingcap/check"
 	"github.com/pingcap/errors"
 	"github.com/pingcap/tidb-binlog/pkg/etcd"
@@ -195,7 +198,11 @@ func (s *mockDrainerServer) Start() (*grpc.Server, error) {
 	}
 	gs := grpc.NewServer()
 	binlog.RegisterCisternServer(gs, s)
-	go gs.Serve(lis)
+	go func() {
+		if err := gs.Serve(lis); err != nil {
+			log.Error("Unexpected exit of gRPC server", zap.Error(err))
+		}
+	}()
 	return gs, nil
 }
 
--- a/pump/server.go
+++ b/pump/server.go
@@ -400,7 +400,11 @@ func (s *Server) Start() error {
 	)
 
 	httpL := m.Match(cmux.HTTP1Fast())
-	go s.gs.Serve(grpcL)
+	go func() {
+		if err := s.gs.Serve(grpcL); err != nil {
+			log.Error("Unexpected exit of gRPC server", zap.Error(err))
+		}
+	}()
 
 	router := mux.NewRouter()
 	router.HandleFunc("/status", s.Status).Methods("GET")
@@ -656,7 +660,9 @@ func (s *Server) startMetrics() {
 func (s *Server) AllDrainers(w http.ResponseWriter, r *http.Request) {
 	node, ok := s.node.(*pumpNode)
 	if !ok {
-		json.NewEncoder(w).Encode("can't provide service")
+		if err := json.NewEncoder(w).Encode("can't provide service"); err != nil {
+			log.Error("Failed to encode msg", zap.Error(err))
+		}
 		return
 	}
 
@@ -665,7 +671,9 @@ func (s *Server) AllDrainers(w http.ResponseWriter, r *http.Request) {
 		log.Error("get pumps failed", zap.Error(err))
 	}
 
-	json.NewEncoder(w).Encode(pumps)
+	if err := json.NewEncoder(w).Encode(pumps); err != nil {
+		log.Error("Failed to encode pumps", zap.Error(err), zap.Any("pumps", pumps))
+	}
 }
 
 // Status exposes pumps' status to HTTP handler.
@@ -776,12 +784,18 @@ func (s *Server) ApplyAction(w http.ResponseWriter, r *http.Request) {
 	log.Info("receive action", zap.String("nodeID", nodeID), zap.String("action", action))
 
 	if nodeID != s.node.NodeStatus().NodeID {
-		rd.JSON(w, http.StatusOK, util.ErrResponsef("invalide nodeID %s, this pump's nodeID is %s", nodeID, s.node.NodeStatus().NodeID))
+		err := rd.JSON(w, http.StatusOK, util.ErrResponsef("invalide nodeID %s, this pump's nodeID is %s", nodeID, s.node.NodeStatus().NodeID))
+		if err != nil {
+			log.Error("Failed to render JSON response", zap.Error(err))
+		}
 		return
 	}
 
 	if s.node.NodeStatus().State != node.Online {
-		rd.JSON(w, http.StatusOK, util.ErrResponsef("this pump's state is %s, apply %s failed!", s.node.NodeStatus().State, action))
+		err := rd.JSON(w, http.StatusOK, util.ErrResponsef("this pump's state is %s, apply %s failed!", s.node.NodeStatus().State, action))
+		if err != nil {
+			log.Error("Failed to render JSON response", zap.Error(err))
+		}
 		return
 	}
 
@@ -793,12 +807,18 @@ func (s *Server) ApplyAction(w http.ResponseWriter, r *http.Request) {
 		log.Info("pump's state change to closing", zap.String("nodeID", nodeID))
 		s.node.NodeStatus().State = node.Closing
 	default:
-		rd.JSON(w, http.StatusOK, util.ErrResponsef("invalide action %s", action))
+		err := rd.JSON(w, http.StatusOK, util.ErrResponsef("invalide action %s", action))
+		if err != nil {
+			log.Error("Failed to render JSON response", zap.Error(err))
+		}
 		return
 	}
 
 	go s.Close()
-	rd.JSON(w, http.StatusOK, util.SuccessResponse(fmt.Sprintf("apply action %s success!", action), nil))
+	err := rd.JSON(w, http.StatusOK, util.SuccessResponse(fmt.Sprintf("apply action %s success!", action), nil))
+	if err != nil {
+		log.Error("Failed to render JSON response", zap.Error(err))
+	}
 }
 
 var utilGetTSO = util.GetTSO
--- a/pump/server_test.go
+++ b/pump/server_test.go
@@ -475,10 +475,10 @@ func (s *gcBinlogFileSuite) TestShouldGCMinDrainerTSO(c *C) {
 	outAlertGCMS := millisecond + (earlyAlertGC+10*time.Minute).Nanoseconds()/1000/1000
 	outAlertGCTS := int64(oracle.EncodeTSO(outAlertGCMS))
 
-	registry.UpdateNode(ctx, "drainers/1", &node.Status{MaxCommitTS: inAlertGCTS, State: node.Online})
-	registry.UpdateNode(ctx, "drainers/2", &node.Status{MaxCommitTS: 1002, State: node.Online})
+	mustUpdateNode(ctx, registry, "drainers/1", &node.Status{MaxCommitTS: inAlertGCTS, State: node.Online})
+	mustUpdateNode(ctx, registry, "drainers/2", &node.Status{MaxCommitTS: 1002, State: node.Online})
 	// drainers/3 is set to be offline, so its MaxCommitTS is expected to be ignored
-	registry.UpdateNode(ctx, "drainers/3", &node.Status{MaxCommitTS: outAlertGCTS, State: node.Offline})
+	mustUpdateNode(ctx, registry, "drainers/3", &node.Status{MaxCommitTS: outAlertGCTS, State: node.Offline})
 
 	// Set a shorter interval because we don't really want to wait 1 hour
 	origInterval := gcInterval
@@ -497,6 +497,12 @@ func (s *gcBinlogFileSuite) TestShouldGCMinDrainerTSO(c *C) {
 
 	c.Assert(storage.gcTS, GreaterEqual, gcTS)
 	// todo: add in and out of alert test while binlog has failpoint
+}
+
+func mustUpdateNode(pctx context.Context, r *node.EtcdRegistry, prefix string, status *node.Status) {
+	if err := r.UpdateNode(pctx, prefix, status); err != nil {
+		panic(err)
+	}
 }
 
 type waitCommitTSSuite struct{}
@@ -745,7 +751,11 @@ func (s *startServerSuite) TestStartPumpServer(c *C) {
 		close(sig)
 		p.Close()
 	}()
-	go p.Start()
+	go func() {
+		if err := p.Start(); err != nil {
+			c.Logf("Pump server stopped in error: %v", err)
+		}
+	}()
 
 	// wait until the server is online
 	timeEnd := time.After(5 * time.Second)
--- a/pump/storage/log.go
+++ b/pump/storage/log.go
@@ -369,7 +369,11 @@ func seekToNextRecord(reader *bufio.Reader) (bytes int, err error) {
 			return
 		}
 
-		reader.Discard(1)
+		if _, err = reader.Discard(1); err != nil {
+			// If we reach here, we've already successfully called `Peek(4)`
+			// and `Discard(1)` should not fail.
+			panic(err)
+		}
 		bytes++
 	}
 }


pkg/loader/load_test.go
```diff
@@ -410,104 +412,116 @@ func outputTxnInTime(c *check.C, output chan *Txn, timeLimit time.Duration) *Txn
...
-	}
-	c.Assert(output, check.HasLen, 4)
-	// Next txn should be blocked
-	select {
-	case input <- txn:
-		c.Fatal("txnManager doesn't block the txn when room is not enough")
-	default:
-	}
-	c.Assert(output, check.HasLen, 4)
+	//input := make(chan *Txn)
```
siddontang: should remove the comment lines.

suzaku:
Thanks for your PR, we'll approve it if it gets at least 2 LGTMs.

drainer/syncer.go
```diff
@@ -14,6 +14,7 @@
 package drainer
 
 import (
+	"github.com/smartystreets-prototypes/go-disruptor"
```
suzaku: Please use `goimports`.

drainer/syncer.go
```diff
@@ -469,11 +498,20 @@ func isIgnoreTxnCommitTS(ignoreTxnCommitTS []int64, ts int64) bool {
...
 	case <-s.shutdown:
-	case s.input <- b:
-		log.Debug("receive publish binlog item", zap.Stringer("item", b))
+	default:
 	}
+	//select {
+	//case <-s.shutdown:
+	//case s.input <- b:
+	//	log.Debug("receive publish binlog item", zap.Stringer("item", b))
+	//}
```
suzaku: Please remove the commented code.

pkg/loader/bench_test.go
```diff
@@ -176,14 +176,15 @@ func (r *runner) close() {
 func createTable(db *sql.DB, loader Loader) error {
 	sql := "create table test1(id int primary key, a1 int)"
 	// sql = "create table test1(id int, a1 int, UNIQUE KEY `id` (`id`))"
-	loader.Input() <- NewDDLTxn("test", "test1", sql)
+	//loader.Input() <- NewDDLTxn("test", "test1", sql)
+	loader.Publish(NewDDLTxn("test", "test1", sql))
```
suzaku: Could you please paste the output of the `loader/bench_test.go` before and after the change?

suzaku:
It seems there's something wrong with the Kafka tests, please take a look.

You may run `make test` to run the unit tests.

billhcmus:
> It seems there's something wrong with the Kafka tests, please take a look.
> 
> You may run `make test` to run the unit tests.

Ok, thank you, at the moment I only running TiDB Binlog in case 2 TiDB cluster.  Let me take a look at the case of Kafka.

pkg/loader/bench_test.go
billhcmus: Ok, I will add more unit-test about disruptors.

suzaku: Cool, correctness comes before performance.

suzaku:
> > It seems there's something wrong with the Kafka tests, please take a look.
> > You may run `make test` to run the unit tests.
> 
> Ok, thank you, at the moment I only running TiDB Binlog in case 2 TiDB cluster. Let me take a look at the case of Kafka.

If you want to run integration tests in your development environment, you can read this [document](https://github.com/pingcap/tidb-binlog/blob/master/tests/README.md).
Note that if you only want to run one type of them you can specify it with the `do_cases` variable in `tests/run.sh`.

leoppro:
Hi, @billhcmus. Thanks for your contribution. 
We researched the [go-disruptor](https://github.com/smartystreets-prototypes/go-disruptor) library. There are two problems here:
1. This library is still in the alpha version. The code is currently experimental and is not recommended for production environments. We are worried about the stability and correctness of `go-disruptor`.
2. We tried to run the benchmarks to verify the performance of `go-disruptor`. It is confusingly that we got two distinct results in different ways.
```
go test -benchmem -run=^$ github.com/smartystreets-prototypes/go-disruptor/benchmarks -bench . -test.v

goos: linux
goarch: amd64
pkg: github.com/smartystreets-prototypes/go-disruptor/benchmarks
BenchmarkChannelBlockingOneGoroutine-2                           	18273165	        65.4 ns/op	       0 B/op	       0 allocs/op
BenchmarkChannelBlockingTwoGoroutines-2                          	16237992	        62.7 ns/op	       0 B/op	       0 allocs/op
BenchmarkChannelBlockingThreeGoroutinesWithContendedWrite-2      	 8818279	       136 ns/op	       0 B/op	       0 allocs/op
BenchmarkChannelNonBlockingOneGoroutine-2                        	65434704	        22.5 ns/op	       0 B/op	       0 allocs/op
BenchmarkChannelNonBlockingTwoGoroutines-2                       	100000000	        24.7 ns/op	       0 B/op	       0 allocs/op
BenchmarkChannelNonBlockingThreeGoroutinesWithContendedWrite-2   	100000000	        91.2 ns/op	       0 B/op	       0 allocs/op
BenchmarkCompositeBarrierRead-2                                  	100000000	        37.1 ns/op	       0 B/op	       0 allocs/op
BenchmarkCursorStore-2                                           	213080528	         5.29 ns/op	       0 B/op	       0 allocs/op
BenchmarkCursorLoad-2                                            	1000000000	         0.453 ns/op	       0 B/op	       0 allocs/op
BenchmarkCursorLoadAsBarrier-2                                   	608153762	         1.89 ns/op	       0 B/op	       0 allocs/op
BenchmarkWriterReserve-2                                         	   14913	     75540 ns/op	       0 B/op	       0 allocs/op
BenchmarkWriterNextWrapPoint-2                                   	64056978	        18.2 ns/op	       0 B/op	       0 allocs/op
BenchmarkWriterCommit-2                                          	178066047	         5.99 ns/op	       0 B/op	       0 allocs/op
BenchmarkWriterReserveOneSingleConsumer-2                        	  895417	      1551 ns/op	       0 B/op	       0 allocs/op
BenchmarkWriterReserveManySingleConsumer-2                       	  630859	      1755 ns/op	       0 B/op	       0 allocs/op
BenchmarkWriterReserveOneMultipleConsumers-2                     	  758583	      1831 ns/op	       0 B/op	       0 allocs/op
BenchmarkWriterReserveManyMultipleConsumers-2                    	  701175	      1780 ns/op	       0 B/op	       0 allocs/op
PASS
coverage: [no statements]
ok  	github.com/smartystreets-prototypes/go-disruptor/benchmarks	210.411s
Success: Benchmarks passed.
```

```
go test -benchmem -run=^$ github.com/smartystreets-prototypes/go-disruptor/benchmarks -bench ^(BenchmarkWriterReserve|BenchmarkWriterNextWrapPoint|BenchmarkWriterCommit|BenchmarkWriterReserveOneSingleConsumer|BenchmarkWriterReserveManySingleConsumer|BenchmarkWriterReserveOneMultipleConsumers|BenchmarkWriterReserveManyMultipleConsumers)$ -test.v

goos: linux
goarch: amd64
pkg: github.com/smartystreets-prototypes/go-disruptor/benchmarks
BenchmarkWriterReserve-2                        	147415915	         7.70 ns/op	       0 B/op	       0 allocs/op
BenchmarkWriterNextWrapPoint-2                  	531537692	         2.33 ns/op	       0 B/op	       0 allocs/op
BenchmarkWriterCommit-2                         	199096694	         5.93 ns/op	       0 B/op	       0 allocs/op
BenchmarkWriterReserveOneSingleConsumer-2       	122751121	         9.67 ns/op	       0 B/op	       0 allocs/op
BenchmarkWriterReserveManySingleConsumer-2      	409466360	         2.87 ns/op	       0 B/op	       0 allocs/op
BenchmarkWriterReserveOneMultipleConsumers-2    	111514148	        10.4 ns/op	       0 B/op	       0 allocs/op
BenchmarkWriterReserveManyMultipleConsumers-2   	344804828	         3.41 ns/op	       0 B/op	       0 allocs/op
PASS
ok  	github.com/smartystreets-prototypes/go-disruptor/benchmarks	11.935s
Success: Benchmarks passed.
```
It seem like `go-disruptor` has serious performance regressions under certain circumstances.

I'm afraid that we can't import this library until the above problems are resolved.

pkg/loader/bench_test.go
billhcmus: Here is the result after run bench_test.go:

|                        | Before                                    | After                                     |   |   |
|-----------------------:|-------------------------------------------|-------------------------------------------|---|---|
| BenchmarkInsertMerge   | 1282 903629 ns/op 2081 B/op 47 allocs/op  | 1488 743148 ns/op 2025 B/op 46 allocs/op  |   |   |
| BenchmarkInsertNoMerge | 1096 1237625 ns/op 2086 B/op 47 allocs/op | 1483 932181 ns/op 1997 B/op 46 allocs/op  |   |   |
| BenchmarkUpdateMerge   | 682 1811123 ns/op 4232 B/op 100 allocs/op | 561 1914113 ns/op 4251 B/op 100 allocs/op |   |   |
| BenchmarkUpdateNoMerge | 546 1905920 ns/op 4232 B/op 100 allocs/op | 684 1674532 ns/op 4131 B/op 99 allocs/op  |   |   |
| BenchmarkDeleteMerge   | 658 1851527 ns/op 3351 B/op 81 allocs/op  | 618 1870392 ns/op 3307 B/op 81 allocs/op  |   |   |
| BenchmarkDeleteNoMerge | 640 1947590 ns/op 3310 B/op 81 allocs/op  | 596 1862614 ns/op 3301 B/op 81 allocs/op  |   |   |

billhcmus: Original format when I run like this:

```
go test -benchmem -run=XXX -bench=.
[2019/12/09 18:13:23.593 +07:00] [INFO] [load.go:360] ["exec ddl success"] [sql="drop table if exists test1"]
[2019/12/09 18:13:23.630 +07:00] [INFO] [load.go:360] ["exec ddl success"] [sql="create table test1(id int primary key, a1 int)"]
[2019/12/09 18:13:23.630 +07:00] [INFO] [load.go:264] ["refresh table info"] [schema=test] [table=test1]
goos: darwin
goarch: amd64
pkg: github.com/pingcap/tidb-binlog/pkg/loader
BenchmarkDeleteMerge-4     	[2019/12/09 18:13:23.678 +07:00] [INFO] [load.go:360] ["exec ddl success"] [sql="drop table if exists test1"]
[2019/12/09 18:13:23.698 +07:00] [INFO] [load.go:360] ["exec ddl success"] [sql="create table test1(id int primary key, a1 int)"]
[2019/12/09 18:13:23.698 +07:00] [INFO] [load.go:264] ["refresh table info"] [schema=test] [table=test1]
[2019/12/09 18:13:23.896 +07:00] [INFO] [load.go:360] ["exec ddl success"] [sql="drop table if exists test1"]
[2019/12/09 18:13:23.918 +07:00] [INFO] [load.go:360] ["exec ddl success"] [sql="create table test1(id int primary key, a1 int)"]
[2019/12/09 18:13:23.918 +07:00] [INFO] [load.go:264] ["refresh table info"] [schema=test] [table=test1]
[2019/12/09 18:13:24.145 +07:00] [INFO] [load.go:360] ["exec ddl success"] [sql="drop table if exists test1"]
[2019/12/09 18:13:24.168 +07:00] [INFO] [load.go:360] ["exec ddl success"] [sql="create table test1(id int primary key, a1 int)"]
[2019/12/09 18:13:24.168 +07:00] [INFO] [load.go:264] ["refresh table info"] [schema=test] [table=test1]
[2019/12/09 18:13:24.781 +07:00] [INFO] [load.go:360] ["exec ddl success"] [sql="drop table if exists test1"]
[2019/12/09 18:13:24.805 +07:00] [INFO] [load.go:360] ["exec ddl success"] [sql="create table test1(id int primary key, a1 int)"]
[2019/12/09 18:13:24.805 +07:00] [INFO] [load.go:264] ["refresh table info"] [schema=test] [table=test1]
     618	   1870392 ns/op	    3307 B/op	      81 allocs/op
[2019/12/09 18:13:25.948 +07:00] [INFO] [load.go:360] ["exec ddl success"] [sql="drop table if exists test1"]
[2019/12/09 18:13:25.972 +07:00] [INFO] [load.go:360] ["exec ddl success"] [sql="create table test1(id int primary key, a1 int)"]
[2019/12/09 18:13:25.972 +07:00] [INFO] [load.go:264] ["refresh table info"] [schema=test] [table=test1]
BenchmarkDeleteNoMerge-4   	[2019/12/09 18:13:26.031 +07:00] [INFO] [load.go:360] ["exec ddl success"] [sql="drop table if exists test1"]
[2019/12/09 18:13:26.053 +07:00] [INFO] [load.go:360] ["exec ddl success"] [sql="create table test1(id int primary key, a1 int)"]
[2019/12/09 18:13:26.053 +07:00] [INFO] [load.go:264] ["refresh table info"] [schema=test] [table=test1]
[2019/12/09 18:13:26.187 +07:00] [INFO] [load.go:360] ["exec ddl success"] [sql="drop table if exists test1"]
[2019/12/09 18:13:26.204 +07:00] [INFO] [load.go:360] ["exec ddl success"] [sql="create table test1(id int primary key, a1 int)"]
[2019/12/09 18:13:26.204 +07:00] [INFO] [load.go:264] ["refresh table info"] [schema=test] [table=test1]
[2019/12/09 18:13:26.478 +07:00] [INFO] [load.go:360] ["exec ddl success"] [sql="drop table if exists test1"]
[2019/12/09 18:13:26.500 +07:00] [INFO] [load.go:360] ["exec ddl success"] [sql="create table test1(id int primary key, a1 int)"]
[2019/12/09 18:13:26.500 +07:00] [INFO] [load.go:264] ["refresh table info"] [schema=test] [table=test1]
[2019/12/09 18:13:27.323 +07:00] [INFO] [load.go:360] ["exec ddl success"] [sql="drop table if exists test1"]
[2019/12/09 18:13:27.350 +07:00] [INFO] [load.go:360] ["exec ddl success"] [sql="create table test1(id int primary key, a1 int)"]
[2019/12/09 18:13:27.350 +07:00] [INFO] [load.go:264] ["refresh table info"] [schema=test] [table=test1]
     596	   1862614 ns/op	    3301 B/op	      81 allocs/op
```

billhcmus:
> Hi, @billhcmus. Thanks for your contribution.
> We researched the [go-disruptor](https://github.com/smartystreets-prototypes/go-disruptor) library. There are two problems here:
> 
> 1. This library is still in the alpha version. The code is currently experimental and is not recommended for production environments. We are worried about the stability and correctness of `go-disruptor`.
> 2. We tried to run the benchmarks to verify the performance of `go-disruptor`. It is confusingly that we got two distinct results in different ways.
> 
> ```
> go test -benchmem -run=^$ github.com/smartystreets-prototypes/go-disruptor/benchmarks -bench . -test.v
> 
> goos: linux
> goarch: amd64
> pkg: github.com/smartystreets-prototypes/go-disruptor/benchmarks
> BenchmarkChannelBlockingOneGoroutine-2                           	18273165	        65.4 ns/op	       0 B/op	       0 allocs/op
> BenchmarkChannelBlockingTwoGoroutines-2                          	16237992	        62.7 ns/op	       0 B/op	       0 allocs/op
> BenchmarkChannelBlockingThreeGoroutinesWithContendedWrite-2      	 8818279	       136 ns/op	       0 B/op	       0 allocs/op
> BenchmarkChannelNonBlockingOneGoroutine-2                        	65434704	        22.5 ns/op	       0 B/op	       0 allocs/op
> BenchmarkChannelNonBlockingTwoGoroutines-2                       	100000000	        24.7 ns/op	       0 B/op	       0 allocs/op
> BenchmarkChannelNonBlockingThreeGoroutinesWithContendedWrite-2   	100000000	        91.2 ns/op	       0 B/op	       0 allocs/op
> BenchmarkCompositeBarrierRead-2                                  	100000000	        37.1 ns/op	       0 B/op	       0 allocs/op
> BenchmarkCursorStore-2                                           	213080528	         5.29 ns/op	       0 B/op	       0 allocs/op
> BenchmarkCursorLoad-2                                            	1000000000	         0.453 ns/op	       0 B/op	       0 allocs/op
> BenchmarkCursorLoadAsBarrier-2                                   	608153762	         1.89 ns/op	       0 B/op	       0 allocs/op
> BenchmarkWriterReserve-2                                         	   14913	     75540 ns/op	       0 B/op	       0 allocs/op
> BenchmarkWriterNextWrapPoint-2                                   	64056978	        18.2 ns/op	       0 B/op	       0 allocs/op
> BenchmarkWriterCommit-2                                          	178066047	         5.99 ns/op	       0 B/op	       0 allocs/op
> BenchmarkWriterReserveOneSingleConsumer-2                        	  895417	      1551 ns/op	       0 B/op	       0 allocs/op
> BenchmarkWriterReserveManySingleConsumer-2                       	  630859	      1755 ns/op	       0 B/op	       0 allocs/op
> BenchmarkWriterReserveOneMultipleConsumers-2                     	  758583	      1831 ns/op	       0 B/op	       0 allocs/op
> BenchmarkWriterReserveManyMultipleConsumers-2                    	  701175	      1780 ns/op	       0 B/op	       0 allocs/op
> PASS
> coverage: [no statements]
> ok  	github.com/smartystreets-prototypes/go-disruptor/benchmarks	210.411s
> Success: Benchmarks passed.
> ```
> 
> ```
> go test -benchmem -run=^$ github.com/smartystreets-prototypes/go-disruptor/benchmarks -bench ^(BenchmarkWriterReserve|BenchmarkWriterNextWrapPoint|BenchmarkWriterCommit|BenchmarkWriterReserveOneSingleConsumer|BenchmarkWriterReserveManySingleConsumer|BenchmarkWriterReserveOneMultipleConsumers|BenchmarkWriterReserveManyMultipleConsumers)$ -test.v
> 
> goos: linux
> goarch: amd64
> pkg: github.com/smartystreets-prototypes/go-disruptor/benchmarks
> BenchmarkWriterReserve-2                        	147415915	         7.70 ns/op	       0 B/op	       0 allocs/op
> BenchmarkWriterNextWrapPoint-2                  	531537692	         2.33 ns/op	       0 B/op	       0 allocs/op
> BenchmarkWriterCommit-2                         	199096694	         5.93 ns/op	       0 B/op	       0 allocs/op
> BenchmarkWriterReserveOneSingleConsumer-2       	122751121	         9.67 ns/op	       0 B/op	       0 allocs/op
> BenchmarkWriterReserveManySingleConsumer-2      	409466360	         2.87 ns/op	       0 B/op	       0 allocs/op
> BenchmarkWriterReserveOneMultipleConsumers-2    	111514148	        10.4 ns/op	       0 B/op	       0 allocs/op
> BenchmarkWriterReserveManyMultipleConsumers-2   	344804828	         3.41 ns/op	       0 B/op	       0 allocs/op
> PASS
> ok  	github.com/smartystreets-prototypes/go-disruptor/benchmarks	11.935s
> Success: Benchmarks passed.
> ```
> 
> It seem like `go-disruptor` has serious performance regressions under certain circumstances.
> 
> I'm afraid that we can't import this library until the above problems are resolved.

Thank you for your reply. I think we shouldn't run the benchmark interleaved go channel and disruptor. 
If separated this, disruptor always efficiently than go channel.
I total agree with you about this alpha version of disruptor and this not ready for the production environments.
- But with this result:


We can consider that approach. (We can deep research and optimize go-disruptor library, for example).

suzaku:
> > Hi, @billhcmus. Thanks for your contribution.
> > We researched the [go-disruptor](https://github.com/smartystreets-prototypes/go-disruptor) library. There are two problems here:
> > 
> > 1. This library is still in the alpha version. The code is currently experimental and is not recommended for production environments. We are worried about the stability and correctness of `go-disruptor`.
> > 2. We tried to run the benchmarks to verify the performance of `go-disruptor`. It is confusingly that we got two distinct results in different ways.
> > 
> > ```
> > go test -benchmem -run=^$ github.com/smartystreets-prototypes/go-disruptor/benchmarks -bench . -test.v
> > 
> > goos: linux
> > goarch: amd64
> > pkg: github.com/smartystreets-prototypes/go-disruptor/benchmarks
> > BenchmarkChannelBlockingOneGoroutine-2                           	18273165	        65.4 ns/op	       0 B/op	       0 allocs/op
> > BenchmarkChannelBlockingTwoGoroutines-2                          	16237992	        62.7 ns/op	       0 B/op	       0 allocs/op
> > BenchmarkChannelBlockingThreeGoroutinesWithContendedWrite-2      	 8818279	       136 ns/op	       0 B/op	       0 allocs/op
> > BenchmarkChannelNonBlockingOneGoroutine-2                        	65434704	        22.5 ns/op	       0 B/op	       0 allocs/op
> > BenchmarkChannelNonBlockingTwoGoroutines-2                       	100000000	        24.7 ns/op	       0 B/op	       0 allocs/op
> > BenchmarkChannelNonBlockingThreeGoroutinesWithContendedWrite-2   	100000000	        91.2 ns/op	       0 B/op	       0 allocs/op
> > BenchmarkCompositeBarrierRead-2                                  	100000000	        37.1 ns/op	       0 B/op	       0 allocs/op
> > BenchmarkCursorStore-2                                           	213080528	         5.29 ns/op	       0 B/op	       0 allocs/op
> > BenchmarkCursorLoad-2                                            	1000000000	         0.453 ns/op	       0 B/op	       0 allocs/op
> > BenchmarkCursorLoadAsBarrier-2                                   	608153762	         1.89 ns/op	       0 B/op	       0 allocs/op
> > BenchmarkWriterReserve-2                                         	   14913	     75540 ns/op	       0 B/op	       0 allocs/op
> > BenchmarkWriterNextWrapPoint-2                                   	64056978	        18.2 ns/op	       0 B/op	       0 allocs/op
> > BenchmarkWriterCommit-2                                          	178066047	         5.99 ns/op	       0 B/op	       0 allocs/op
> > BenchmarkWriterReserveOneSingleConsumer-2                        	  895417	      1551 ns/op	       0 B/op	       0 allocs/op
> > BenchmarkWriterReserveManySingleConsumer-2                       	  630859	      1755 ns/op	       0 B/op	       0 allocs/op
> > BenchmarkWriterReserveOneMultipleConsumers-2                     	  758583	      1831 ns/op	       0 B/op	       0 allocs/op
> > BenchmarkWriterReserveManyMultipleConsumers-2                    	  701175	      1780 ns/op	       0 B/op	       0 allocs/op
> > PASS
> > coverage: [no statements]
> > ok  	github.com/smartystreets-prototypes/go-disruptor/benchmarks	210.411s
> > Success: Benchmarks passed.
> > ```
> > 
> > 
> > ```
> > go test -benchmem -run=^$ github.com/smartystreets-prototypes/go-disruptor/benchmarks -bench ^(BenchmarkWriterReserve|BenchmarkWriterNextWrapPoint|BenchmarkWriterCommit|BenchmarkWriterReserveOneSingleConsumer|BenchmarkWriterReserveManySingleConsumer|BenchmarkWriterReserveOneMultipleConsumers|BenchmarkWriterReserveManyMultipleConsumers)$ -test.v
> > 
> > goos: linux
> > goarch: amd64
> > pkg: github.com/smartystreets-prototypes/go-disruptor/benchmarks
> > BenchmarkWriterReserve-2                        	147415915	         7.70 ns/op	       0 B/op	       0 allocs/op
> > BenchmarkWriterNextWrapPoint-2                  	531537692	         2.33 ns/op	       0 B/op	       0 allocs/op
> > BenchmarkWriterCommit-2                         	199096694	         5.93 ns/op	       0 B/op	       0 allocs/op
> > BenchmarkWriterReserveOneSingleConsumer-2       	122751121	         9.67 ns/op	       0 B/op	       0 allocs/op
> > BenchmarkWriterReserveManySingleConsumer-2      	409466360	         2.87 ns/op	       0 B/op	       0 allocs/op
> > BenchmarkWriterReserveOneMultipleConsumers-2    	111514148	        10.4 ns/op	       0 B/op	       0 allocs/op
> > BenchmarkWriterReserveManyMultipleConsumers-2   	344804828	         3.41 ns/op	       0 B/op	       0 allocs/op
> > PASS
> > ok  	github.com/smartystreets-prototypes/go-disruptor/benchmarks	11.935s
> > Success: Benchmarks passed.
> > ```
> > 
> > 
> > It seem like `go-disruptor` has serious performance regressions under certain circumstances.
> > I'm afraid that we can't import this library until the above problems are resolved.
> 
> Thank you for your reply. I think we shouldn't run the benchmark interleaved go channel and disruptor.
> If separated this, disruptor always efficiently than go channel.
> I total agree with you about this alpha version of disruptor and this not ready for the production environments.
> 
> * But with this result:
> 
> 
> We can consider that approach. (We can deep research and optimize go-disruptor library, for example).

Sure, the improvement looks attractive.  Things I think we should focus on  now:

1. (correctness) Fix the broken unit tests and integration tests
1. (performance) Prove that go-disruptor is indeed much faster than using channel directly (so that it's worth making the code more complex to introduce it)
1. (correctness) Look into more testing and code reading of the go-disruptor project

billhcmus:
> Sure, the improvement looks attractive. Things I think we should focus on now:
> 
> 1. (correctness) Fix the broken unit tests and integration tests
> 2. (performance) Prove that go-disruptor is indeed much faster than using channel directly (so that it's worth making the code more complex to introduce it)
> 3. (correctness) Look into more testing and code reading of the go-disruptor project

You're right, I will take a look about the correctness (unit tests and integration tests) first.

july2993:
Can you show both `drainer event` dashboard and the `checkpoint delay` together? also the origin drainer version.
the `checkpoint delay` you show change move up or down to fast, I think it's abnormal and maybe just caused by how we calculate the metrics like don't handle the success event ASAP.


If the delay is caused by we blocking at some `chan` and doing nothing when push the entry into `successTxn`(no one consume it ASAP) in pkg/Loader after writing a batch entries we can improve it directly like use a larger buffer chan and affect is same as you use go-disruptor here. I don't think the performance of `chan` is a problem here.

The txnManager is used to control how much DML entry to buffer because one Txn may contain huge dml entries or only one, it 's unavailable to limit the buffer data(or memory) by just setting a buffer size for Txn.

billhcmus:
> Can you show both `drainer event` dashboard and the `checkpoint delay` together? also the origin drainer version.
> the `checkpoint delay` you show change move up or down to fast, I think it's abnormal and maybe just caused by how we calculate the metrics like don't handle the success event ASAP.
> 
> If the delay is caused by we blocking at some `chan` and doing nothing when push the entry into `successTxn`(no one consume it ASAP) in pkg/Loader after writing a batch entries we can improve it directly like use a larger buffer chan and affect is same as you use go-disruptor here. I don't think the performance of `chan` is a problem here.
> 
> The txnManager is used to control how much DML entry to buffer because one Txn may contain huge dml entries or only one, it 's unavailable to limit the buffer data(or memory) by just setting a buffer size for Txn.

Yes, you're right, after change cacheChan size in txnManager to 2^16 and re-benchmark tidb-binlog, the delay decrease absolutely. So the problem not in `chan`, but in the `chan` size. So the question is: What is the efficient value of the cacheChane? Default it's 1024

billhcmus:
So now the go-disruptor is not necessary

july2993:
> > Can you show both `drainer event` dashboard and the `checkpoint delay` together? also the origin drainer version.
> > the `checkpoint delay` you show change move up or down to fast, I think it's abnormal and maybe just caused by how we calculate the metrics like don't handle the success event ASAP.
> > If the delay is caused by we blocking at some `chan` and doing nothing when push the entry into `successTxn`(no one consume it ASAP) in pkg/Loader after writing a batch entries we can improve it directly like use a larger buffer chan and affect is same as you use go-disruptor here. I don't think the performance of `chan` is a problem here.
> > The txnManager is used to control how much DML entry to buffer because one Txn may contain huge dml entries or only one, it 's unavailable to limit the buffer data(or memory) by just setting a buffer size for Txn.
> 
> Yes, you're right, after change cacheChan size in txnManager to 2^16 and re-benchmark tidb-binlog, the delay decrease absolutely. So the problem not in `chan`, but in the `chan` size. So the question is: What is the efficient value of the cacheChane? Default it's 1024

Do you change only the buffer size of `cacheChan` in `newTxnManager`, if yes, this should take no different because it limit 1024(maxCacheSize) dml entries in the `cacheChan`, so the chan should always be no full setting to be 1024.

GregoryIan:
Hi @billhcmus , as @july2993 asked , do you  only change the size of `cacheChan` in `newTxnManager`, it's very strange, and what is your current thought on this problem?

billhcmus:
> Do you change only the buffer size of `cacheChan` in `newTxnManager`, if yes, this should take no different(after I reviewing the relate code) because it limit 1024(maxCacheSize) dml entries in the `cacheChan`, so the chan should always be no full setting to be 1024.

- Yep, I change only buffer size.
- But if buffer is full, `input` channel will be blocked.

billhcmus:
> Hi @billhcmus , as @july2993 asked , do you only change the size of `cacheChan` in `newTxnManager`, it's very strange, and what is your current thought on this problem?

I only chan the `cacheChan` size,
I will add metric about the size of the `cacheChan`, I think it were blocked when upstream has high throughput and 1024 is not enough

july2993:
> > Do you change only the buffer size of `cacheChan` in `newTxnManager`, if yes, this should take no different(after I reviewing the relate code) because it limit 1024(maxCacheSize) dml entries in the `cacheChan`, so the chan should always be no full setting to be 1024.
> 
> * Yep, I change only buffer size.
> * But if buffer is full, `input` channel will be blocked.

- In txnManager.run(), it will limit the total dml size in `ret` or `t.cacheChan` at most 1024(the default value of `maxCacheSize`), so it will never be blocked at L696.https://github.com/pingcap/tidb-binlog/blob/628c34a97114088afeb0c6878cb66fec2ba7cb74/pkg/loader/load.go#L696
So only change the `cacheChan` should take no different in theory.

- Note in the origin description you show, the checkpoint delay move up or down too fast, I wonder if it's executing DDL like `add index...` may be slow, that's why I ask for the `drainer event` dashboard metrics too, you can see what it's replicating at that time and the speed for DML event (Insert/Delete/Update). You should compare at some workload and the same downstream system if you make some change.

suzaku:
* `txnManager.cacheChan` is of type `chan *Txn`, it's a buffered channel of `*Txn`s (which corresponds to binlogs);
* Most `*Txn`s are DML transactions, which contains at least 1 `DML`;
* `maxCacheSize` is the maximum number of DMLs allowed in `cacheChan`

Since the size of `cacheChan` and the default value of `maxCacheSize` are both hardcoded to be `1024`, as long as the average number of `DML`s per transaction is more than `1` (this should be the case for most scenarios), the `cacheChan` would not be full because `txnManager` would first block because `maxCacheSize` is reached.

suzaku:
@billhcmus BTW, you can visit the [sig-tidb-binlog](https://app.slack.com/client/TH91JCS4W/CRJKS9HB9) channel on Slack if you have any question about `tidb-binlog`.

billhcmus:
Here is the drainer event metric:

image

billhcmus:
> * `txnManager.cacheChan` is of type `chan *Txn`, it's a buffered channel of `*Txn`s (which corresponds to binlogs);
> * Most `*Txn`s are DML transactions, which contains at least 1 `DML`;
> * `maxCacheSize` is the maximum number of DMLs allowed in `cacheChan`
> 
> Since the size of `cacheChan` and the default value of `maxCacheSize` are both hardcoded to be `1024`, as long as the average number of `DML`s per transaction is more than `1` (this should be the case for most scenarios), the `cacheChan` would not be full because `txnManager` would first block because `maxCacheSize` is reached.

- I know that. But blocked in `txnManager` is  not different from blocked in `cacheChan`, except `cacheChan` is never full.

- `But if buffer is full, input channel will be blocked.`  I use the wrong word, sorry about that

- You can change the `cacheChan` size to `1 << 16` then compare with default `cacheChan` to see the `check point delay` @july2993 . I think it has some impact.

GregoryIan:
@billhcmus yep, do you mean the case is not  blocked in txnManager, but blocked in `cache channel`? 
if so, the case must meet  `txnSize + cachedSize < maxCacheSize(1024)`,  it implies there must be some ddl txn already in `cache channel`, right?

GregoryIan:
the metric picture looks like it’s not what we want. Can you provide the metric picture of drainer event when there is a relatively large delay？

GregoryIan:
@billhcmus and if blocked in txnManager, increasing cache size takes no effect, it is very weird

GregoryIan:
from @july2993 ,  Is the first image showing drainer event  is the one when you change size to be 2^16 but not the one according to when the delay is high? and can you show you origin sysbench bench script, I can do a comparison test myself.  @billhcmus

GregoryIan:
After offline discussion, we find that @billhcmus also had changed `maxCacheSize`, then the above results are as expected. The question now becomes: how do we increase its default value without causing OOM,  `2^16` may be already a suitable value. Can you do more tests based on different workload and observe the memory usage? @billhcmus

billhcmus:
> After offline discussion, we find that @billhcmus also had changed `maxCacheSize`, then the above results are as expected. The question now becomes: how do we increase its default value without causing OOM, `2^16` may be already a suitable value. Can you do more tests based on different workload and observe the memory usage? @billhcmus

It's good to hear from you, ofcourse will do more test with real data, I think we should add this parameter (bufferSize or something like this) to the configuration of the loader.

GregoryIan:
make sense @billhcmus, can you help with it?

billhcmus:
> make sense @billhcmus, can you help with it?

yeah, I will

PR #846, opened 2019-12-12
pump: Add a log for when MVCC is not found for a txn not yet GCed

suzaku:

There's no log for this case.

Add the log.

### Check List 

Tests 

 - Unit test


GregoryIan:
Good Job!  LGTM

GregoryIan:
@suzaku I always think one problem: we hold `tidb` and `tikv`, why can't `tikv` provide a interface tell the   more specific information instead of relying on `maxTxnTimeoutSecond`?

Issue #851, opened 2019-12-18
table not sync


you06:
## Bug Report

1. What did you do?

There is not a way to ensure this problem can be reproduced. Main steps:

* Write data (insert/update/create table) for a while

* Submit all transactions and write a new table immediately

* Repeat the steps before with chaos operations including kill tikv/pump/drainer and network partition between drainer and one of the 3 pump nodes

2. What did you expect to see?

All tables are sync into targetdb.

3. What did you see instead?

From source db

```
mysql> SHOW TABLES;
...
2110 rows in set (0.00 sec)

mysql> DESC t1576547400;
+-------+---------+------+------+---------+-------+
| Field | Type    | Null | Key  | Default | Extra |
+-------+---------+------+------+---------+-------+
| id    | int(11) | YES  |      | NULL    |       |
+-------+---------+------+------+---------+-------+
1 row in set (0.00 sec)
```

From target db

```
mysql> SHOW TABLES;
...
2109 rows in set (0.01 sec)

mysql> DESC t1576547400;
ERROR 1146 (42S02): Table 'sqlsmith.t1576547400' doesn't exist
```

4. Please provide the relate downstream type and version of drainer.
(run `drainer -V` in terminal to get drainer's version)

```
# ./drainer -V
Release Version: v3.0.7
Git Commit Hash: a7b869c024a62eeb7923de4805ab39ec46586e2d
Build TS: 2019-12-04 10:05:55
Go Version: go1.13
Go OS/Arch: linux/amd64
```

leoppro:
fixed : #852

PR #852, opened 2019-12-19
Bug fix: a DDL p-binlog will be ignored when the c-binlog queried from TiKV

leoppro:
The DDL `c-binlog`s sent from TiDB include `ddl query`, `ddl job id`. But the `c-binlog` queried from TiKV doesn't include this info. Unfortunately, pump never read the `ddl query`, `ddl job id` from `p-binlog`. So a DDL `p-binlog` will be ignored when the c-binlog queried from TiKV.
 
fix: #851
Copy the `ddl query`, `ddl job id` from `p-binlog` to `c-binlog`.

### Check List 

Tests 

 - Unit test


suzaku:
Add a test case that could reproduce the bug?

leoppro:
> Add a test case that could reproduce the bug?

fine

PR #856, opened 2019-12-23
tests: Make sure the INSERT is executed before we starts droppling cols

suzaku:

Sometimes the goroutine that drops columns runs before we insert the row with all columns and cause error.

Make sure the insertion happens before updating and column dropping starts.

### Check List 

Tests 

 - Integration test


leoppro:
LGTM

suzaku:
@leoppro Please approve this PR.

PR #857, opened 2019-12-26
Add encrypt password config support

july2993:
also add a command to binlogctl, usage example:
./binlogctl -cmd encrypt -text aaa

Add encrypt password config support

### Check List 

Tests 

 - Unit test

Code changes

Side effects

Related changes


binlogctl/encrypt.go
```diff
@@ -0,0 +1,18 @@
...
+)
+
+// EncryptHandler log the encrypted text if success or return error.
+func EncryptHandler(text string) error {
+	enc, err := encrypt.Encrypt(text)
+	if err != nil {
+		return err
+	}
+
+	log.Info("encrypt text", zap.String("encrypted", string(enc)), zap.String("origin", text))
```
leoppro: Is is fine if we print origin text in log? Maybe secret info will be leaked

pkg/encrypt/encrypt.go
```diff
@@ -0,0 +1,122 @@
...
+	"crypto/cipher"
+	"crypto/rand"
+	"encoding/base64"
+	"encoding/hex"
+
+	"github.com/pingcap/errors"
+)
+
+var (
+	secretKey, _ = hex.DecodeString("a529b7665997f043a30ac8fadcb51d6aa032c226ab5b7750530b12b8c1a16a48")
```
leoppro: hard code here?

july2993: yes, currently works like DM, or you need the get this in some way.

suzaku: Since we are using Ansible, we can use Ansible Vault.

The workflow is like:

1. Store sensitive messages in Ansible Vault
1. Use environment variables that are set in playbook to pass the values
1. When deploying, Ansible will ask the operator for password to decrypt the message
1. In our code we read the sensitive values from an environment variable

july2993: @suzaku @leoppro PTAL 
https://github.com/pingcap/tidb-binlog/pull/857/commits/60a5f622e8d604167e618e1f2a4ec59de5f61c9d
Support set secretkey by env

suzaku: Can we go one step further?

1. We don't do any password en/decryption
1. Introduce several environment variables to pass database passwords, for example, you can use `MYSQL_PWD` to specify the password.

In Linux, only the root and the owner of the process can read the environment variables, so it could be considered safe as long as the system administrator know to keep his/her account safe.

suzaku: Related discussion: https://security.stackexchange.com/questions/83180/secure-database-password-stored-in-an-env-variable

GregoryIan: setting secretkey by enviornment variable is ok now, and I suggest to file an issue for PM: providing a secure and unified encryption way

GregoryIan:
please make sure this PR alive

july2993:
> please make sure this PR alive

there's some thing error about encoding with the latest   tidb/tikv,    @lysu need the ci env to diagnose it.
not relate to this pr, I think you can still review this.

pkg/encrypt/encrypt.go
july2993: @suzaku @leoppro PTAL

PR #860, opened 2019-12-26
Update parser as latest verison tidb use

july2993:
Update parser as latest verison tidb use

Update parser as latest verison tidb use

### Check List 

Tests


leoppro:
LGTM

PR #863, opened 2019-12-31
Avoid the sleep style for doing nothing

july2993:
Fix drainer may just spend most time sleeping.

the work load in Syncer.sync may like this before:
loop {
  1, consume all buffer job in jobChan.
  2, sleep 10ms for doing nothing although we are pushing job into jobChan
in another goroutine.
}

this hurts especial when the buffer size of jobChan is small.
most of the time will just sleeping.

This PR avoid the bad sleep code style, and just keep the same logic.
Note for dest=file/kafka will never `execute` in the maxExecutionWaitTime case

### Check List 

Tests 

 - Integration test
Screen Shot 2019-12-31 at 5 17 04 PM

1. ./bin/drainer -log-file /tmp/tidb_binlog_test/drainer.log -L info -dest-db-type pb  (v2.1.18)
2. ./bin/drainer -log-file /tmp/tidb_binlog_test/drainer.log -L info -dest-db-type pb -c 1 -txn-batch 1 (v2.1.18)
3 ./bin/drainer -log-file /tmp/tidb_binlog_test/drainer.log -L info -dest-db-type pb -c 1 -txn-batch 1 (this pr)

Note the buffer size of jobChan is changed in https://github.com/pingcap/tidb-binlog/pull/735
so some older version with txn-batch = 1 will get a large buffer size and will not be slow.

Code changes

Side effects

 

Related changes


GregoryIan:
LGTM

leoppro:
please fix the CI

drainer/syncer.go
```diff
@@ -323,8 +322,13 @@ func (s *Syncer) sync(executor executor.Executor, jobChan chan *job, executorIdx
...
 		args = args[0:0]
 		commitTSs = commitTSs[0:0]
-		s.lastSyncTime = time.Now()
-		lastSyncTime = time.Now()
+		now := time.Now()
+		s.lastSyncTime = now
+		if !maxExecutionWaitTimer.Stop() {
+			<-maxExecutionWaitTimer.C
+		}
+		maxExecutionWaitTimer.Reset(maxExecutionWaitTime)
```
suzaku: Is it more straight forward if we reset the timer in the `for` loop right before the `select`?

july2993: Onely need to reset when `execute` is called

july2993:
@leoppro PTAL fixed a case where we had consume the timer.C and still try to drain it

drainer/syncer.go
```diff
@@ -323,8 +323,15 @@ func (s *Syncer) sync(executor executor.Executor, jobChan chan *job, executorIdx
 		sqls = sqls[0:0]
 		args = args[0:0]
 		commitTSs = commitTSs[0:0]
-		s.lastSyncTime = time.Now()
-		lastSyncTime = time.Now()
+		now := time.Now()
+		s.lastSyncTime = now
+		if !maxExecutionWaitTimer.Stop() {
+			if !draineTimer {
+				<-maxExecutionWaitTimer.C
```
suzaku:
```go
select {
  case <-maxExecutionWaitTimer.C:
  default:
  }
```

Can we use `select` here to avoid managing the extra boolean variable?

july2993: I don't think we can assume we can trigger the `case <-maxExecutionWaitTimer.C:` absolutely, that means then the event may be pushed into the `chan` by another goroutine.
better make it explicitly when we need to drain it as the docs.

suzaku: Do we need to set `draineTimer` back to `false` in other cases?

july2993: https://github.com/pingcap/tidb-binlog/pull/863/commits/eada8b703002893b86e77a62769f9462e55d83ac

leoppro:
LGTM

july2993:
> LGTM

pls Approve

PR #868, opened 2020-01-07
Add encrypt password config support (#857)

july2993:
pick #857

* Add encrypt password support

also add a command to binlogctl, usage example:
./binlogctl -cmd encrypt -text aaa


WangXiangUSTC:
LGTM

PR #870, opened 2020-01-08
metrics: add grafana and alertmanager scripts

WangXiangUSTC:
add binlog's grafana and alertmanager scripts, include kafka

cp from tidb-ansible


WangXiangUSTC:
@july2993 @suzaku PTAL

metrics/grafana/reparo.json
```diff
@@ -0,0 +1,609 @@
+{
```
july2993: what the hell is this file? I can't find any related metrics, I think it's never used.

WangXiangUSTC: cp from https://github.com/pingcap/tidb-ansible/blob/master/scripts/reparo.json

WangXiangUSTC: seems is uselss, I will delete it

july2993: how about also remove the ones about kafka since it's not related to this repo(like downstream tidb).

WangXiangUSTC: ok

PR #864, opened 2020-01-03
Support two clusters of bidirectional replication.

freemindLi:
### What problem does this PR solve? 
Support two clusters of bidirectional replication.  
A <-> B

### What is changed and how it works?
- surport ddl one-way synchronization
- surport dml Mutual synchronization

### Check List 

Tests 

 - Unit test
 - Integration test
 - Manual test (add detailed scripts or steps below)

Code changes

Side effects

Related changes

 - Need to cherry-pick to the release branch
 - Need to update the documentation
 - Need to be included in the release note

PR #879, opened 2020-01-09
binlog loop back sync

freemindLi:
recreate of #864
In the AA dual activity scenario, there is data write (DML) on both sides. The data of tidb cluster on both sides needs to be synchronized with each other，but avoid loopback synchronization，and ddl is only writed one side.

add three variables in drainer.toml as identification to confirm need sync ddl or not and need set sync mark identification or not and sync identification id to Avoid loopback synchronization
add configuration item
loopback-control (true/false) set mark table identification or not and filter txn by mark table
ddl-sync (true/false) sync ddl to downstream DB or not
channel-id (integer) sync identification id，avoid loopback synchronization

### Check List 

Tests 

 - Unit test
 - Integration test
will add in #882

Related changes


july2993:
/run-all-tests
@freemindLi  you can just add a comment like this to trigger all the tests

pkg/loader/load.go
```diff
@@ -438,8 +451,39 @@ func (s *loaderImpl) execDMLs(dmls []*DML) error {
...
+func (s *loaderImpl) createMarkTableDDL() error {
+	markTableDataBase := loopbacksync.MarkTableName[:strings.Index(loopbacksync.MarkTableName, ".")]
+	createDatabaseSQL := fmt.Sprintf("create database IF NOT EXISTS %s;", markTableDataBase)
+	createDatabase := DDL{SQL: createDatabaseSQL}
+	if s.db == nil || s.loopBackSyncInfo == nil {
+		log.Error("loaderImpl db point object is nil,maybe not create mark table and database or not need creat")
+		return nil
+	}
+	if err1 := s.execDDL(&createDatabase); err1 != nil {
+		if !pkgsql.IgnoreDDLError(err1) {
```
july2993: don't need to check for `IgnoreDDLError` here we have using `IF NOT EXISTS`
`IgnoreDDLError` is used to ignore some error cause we re-execute it at downstream DB.

pkg/loader/load.go
```diff
@@ -438,8 +451,39 @@ func (s *loaderImpl) execDMLs(dmls []*DML) error {
...
+			log.Error("exec failed", zap.String("sql", createDatabase.SQL), zap.Error(err1))
+			return errors.Trace(err1)
+		}
+		log.Warn("ignore ddl", zap.Error(err1), zap.String("ddl", createDatabase.SQL))
+	}
+	sql := createMarkTable()
+	createMarkTable := DDL{Database: markTableDataBase, Table: loopbacksync.MarkTableName, SQL: sql}
+	if err := s.execDDL(&createMarkTable); err != nil {
+		if !pkgsql.IgnoreDDLError(err) {
```
july2993: ditto

pkg/loader/load.go
```diff
@@ -438,8 +451,39 @@ func (s *loaderImpl) execDMLs(dmls []*DML) error {
...
+			return errors.Trace(err)
+		}
+		log.Warn("ignore ddl", zap.Error(err), zap.String("ddl", createMarkTable.SQL))
+	}
+	return nil
+}
+
 // Run will quit when meet any error, or all the txn are drained
 func (s *loaderImpl) Run() error {
+
```
july2993: remove blank line?

pkg/loader/load.go
```diff
@@ -438,8 +451,39 @@ func (s *loaderImpl) execDMLs(dmls []*DML) error {
 	return errors.Trace(err)
 }
 
+func (s *loaderImpl) createMarkTableDDL() error {
+	markTableDataBase := loopbacksync.MarkTableName[:strings.Index(loopbacksync.MarkTableName, ".")]
+	createDatabaseSQL := fmt.Sprintf("create database IF NOT EXISTS %s;", markTableDataBase)
+	createDatabase := DDL{SQL: createDatabaseSQL}
+	if s.db == nil || s.loopBackSyncInfo == nil {
+		log.Error("loaderImpl db point object is nil,maybe not create mark table and database or not need creat")
```
july2993: `s.db` should never be nil.
Do not log an error msg and return nil if `s.loopBackSyncInfo` is nil cause seems it's valid to be nil as default value you set it.

make clear when we need to create the table, if we don't need, should there's no ans error msg.
if we do need, should success to create it or return error and `Loader` failed to work.

drainer/sync/mysql.go
```diff
@@ -15,6 +15,7 @@ package sync
 
 import (
 	"database/sql"
+	"github.com/pingcap/tidb-binlog/drainer/loopbacksync"
```
july2993: can you use goimports to format the code?
It can also help you add or remove the import path automatically.

pkg/loader/model.go
```diff
@@ -184,6 +186,22 @@ func (dml *DML) updateSQL() (sql string, args []interface{}) {
 	return
 }
 
+func updateMarkSQL(columns []string, Values map[string]interface{}) (string, []interface{}) {
+
```
july2993: remove blank line?

drainer/syncer.go
```diff
@@ -439,6 +459,31 @@ ForLoop:
...
+			}
+		}
+	}
+	return false, nil
+}
+
+func loopBackStatus(binlog *pb.Binlog, prewriteValue *pb.PrewriteValue, infoGetter translator.TableInfoGetter, info *loopbacksync.LoopBackSync) (bool, error) {
+	var tableName string
+	var schemaName string
+	tnx, err := translator.TiBinlogToTxn(infoGetter, schemaName, tableName, binlog, prewriteValue)
```
july2993: s/tnx/txn

suzaku:
Please resolve the conflict.

drainer/syncer.go
```diff
@@ -439,6 +459,31 @@ ForLoop:
 	return cerr
 }
 
+func filterMarkDatas(dmls []*loader.DML, info *loopbacksync.LoopBackSync) (bool, error) {
+	for _, dml := range dmls {
+		tableName := dml.Database + "." + dml.Table
+		if strings.EqualFold(tableName, loopbacksync.MarkTableName) {
+			channelID, ok := dml.Values[loopbacksync.ChannelID]
+			if ok {
+				if channelID.(loopbacksync.Channel) == info.ChannelID {
```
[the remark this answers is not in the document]
suzaku: Maybe it's simpler to just use `int64` instead of introducing a new type?

drainer/syncer.go
```diff
@@ -439,6 +459,31 @@ ForLoop:
 	return cerr
 }
 
+func filterMarkDatas(dmls []*loader.DML, info *loopbacksync.LoopBackSync) (bool, error) {
```
suzaku: This function can be called `findLoopBackMark`.

pkg/loader/executor.go
```diff
@@ -50,6 +53,10 @@ func (e *executor) withBatchSize(batchSize int) *executor {
 	return e
 }
 
+func (e *executor) setsyncInfo(info *loopbacksync.LoopBackSync) {
```
suzaku: `setSyncInfo`?

pkg/loader/executor.go
```diff
@@ -102,17 +109,43 @@ func (tx *tx) commit() error {
...
+func (e *executor) updateMark(status int, channel string, tx *tx) error {
+	if e.info == nil {
+		return nil
+	}
+	values := map[string]interface{}{
+		loopbacksync.ChannelID:   e.info.ChannelID,
+		loopbacksync.Val:         status,
+		loopbacksync.ChannelInfo: channel}
+	columns := []string{loopbacksync.ChannelID, loopbacksync.Val, loopbacksync.ChannelInfo}
+	sql, args := updateMarkSQL(columns, values)
```
suzaku: Since the table definition is static, we may not need an extra function for dynamic creation here?
For example, we can just get the `SQL` and `args` ready in this function without using a `map[string]interface{}`.

pkg/loader/load.go
```diff
@@ -438,8 +452,30 @@ func (s *loaderImpl) execDMLs(dmls []*DML) error {
 	return errors.Trace(err)
 }
 
+func (s *loaderImpl) createMarkTableDDL() error {
```
suzaku: It seems the names of the two function `createMarkTableDDL` and `createMarkTable` should be swapped.

cmd/drainer/drainer.toml
```diff
@@ -38,6 +38,15 @@ compressor = ""
...
 txn-batch = 20
 
+# syn ddl to downstream db
+ddl-sync = true
+
+# set mark identification to downstream db mark table
+mark-status = false
+
+# in AA environment, the unique identification of back and forth synchronization prevents loop synchronization
+channel-id = 1
```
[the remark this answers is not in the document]
july2993: channel is ok, it's not a keywork(chan is) and there's no conflict.

freemindLi:
fix code format and fix some bugs；Resolving conflicts

pkg/loader/executor.go
```diff
@@ -102,17 +109,41 @@ func (tx *tx) commit() error {
 	return errors.Trace(err)
 }
 
+func (e *executor) updateMark(status int, channel string, tx *tx) error {
```
july2993: s/status/initValue  ?    or just remove this param, set it zero or one in updateMark.

freemindLi:
1.adjust explanatory note for sync-ddl,loopback-control and channel-id
2.rename stream_info to channel_info
3.delete updateMark function status param
4.adjust column hard code to const reference

pkg/loader/model_test.go
```diff
@@ -232,21 +233,20 @@ func (s *SQLSuite) TestUpdateSQL(c *check.C) {
 }
 
 func (s *SQLSuite) TestUpdateMarkSQL(c *check.C) {
```
july2993: remove this test or use sqlmock to test it, because of this test case test nothing actually,  it just rewrites some codes in the test and can't prove the correction of `updateMark`  (like if we wrongly update updateMark) this test will still pass always.

cmd/drainer/drainer.toml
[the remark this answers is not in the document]
suzaku: Channel may be confusing to maintainers since it's a concept that Go developers used in a daily base. In my opinion it's not a good idea to reuse the name "Channel" when we have other names available.

freemindLi:
1.adjust TestUpdateMarkSQL function logic

pkg/loader/model_test.go
```diff
@@ -232,18 +234,27 @@ func (s *SQLSuite) TestUpdateSQL(c *check.C) {
 	c.Assert(args[1], check.Equals, "pingcap")
 }
 
+//func (s *SQLSuite) TestUpdateMarkSQL(c *check.C) {
```
july2993: remove this comentted line?

freemindLi:
remove comment line

PR #847, opened 2019-12-12
drainer: support relay log for syncer

djshow832:
### What problem does this PR solve? 
Support relay log in Drainer so that the downstream can always be recovered by relay log. Refer to issue #842 for more details.
This is the first part of the feature. This request only supports writing relay log before syncing to the downstream.

### What is changed and how it works?
Add a package `relay`, it supports writing binlog. When `MysqlSyncer` syncs to the downstream, it writes to relay log first. After `MysqlSyncer` successfully writes to the downstream, it purges unused relay log.

### Check List 

Tests 

 - Unit test

Code changes

 - Has exported function/method change
 - Has exported variable/fields change
 - Has interface methods change

Side effects

 - Possible performance regression
 - Increased code complexity

Related changes

 - Need to update the documentation
 - Need to be included in the release note

PR #883, opened 2020-01-14
feed downstream by relay log at startup if need

july2993:
support to recover at startup by relay log. 
follows #847 #849 
- add a status in checkpoint
    - we can learn from the status that there are dirty data written into downstream behind commit ts or not.
- group the config items about relay
    - remove the `relay-read-buf-size`
    - rename `relay-log-size` to `max-file-size`
- make relay.Reader return the origin format  directly
    - this will be more flexible by the user of the reader
- feed downstream by relay log at startup if need.
    - mainly implement in `drainer/relay.go`
    - only fails to init the pd client will try this or just assume the upstream cluster is normal and works like before.

   

### Check List 

Tests 

 - Unit test
- kill -9 drainer & pd while insert data into upstream
- check status is Running
- startup drainer
- check recover and status updated to be Normal

Side effects

Related changes

### DIFF
--- a/drainer/checkpoint/checkpoint.go
+++ b/drainer/checkpoint/checkpoint.go
@@ -61,7 +61,7 @@ func NewCheckPoint(cfg *Config) (CheckPoint, error) {
 	case "mysql", "tidb":
 		cp, err = newMysql(cfg)
 	case "file":
-		cp, err = NewFile(cfg)
+		cp, err = NewFile(cfg.InitialCommitTS, cfg.CheckPointFile)
 	default:
 		err = errors.Errorf("unsupported checkpoint type %s", cfg.CheckpointType)
 	}
--- a/drainer/checkpoint/checkpoint.go
+++ b/drainer/checkpoint/checkpoint.go
@@ -20,8 +20,8 @@ import (
 )
 
 const (
-	// StatusNormal means server quit normally, data <= ts is synced to downstream
-	StatusNormal int = 0
+	// StatusConsistent means server quit normally, data <= ts is synced to downstream
+	StatusConsistent int = 0
 
 	// StatusRunning means server running or quit abnormally, part of data may or may not been synced to downstream
 	StatusRunning int = 1
--- a/drainer/checkpoint/checkpoint.go
+++ b/drainer/checkpoint/checkpoint.go
@@ -19,6 +19,14 @@ import (
 	"go.uber.org/zap"
 )
 
+const (
+	// StatusNormal means server quit normally, data <= ts is synced to downstream
+	StatusNormal int = 0
+
+	// StatusRunning means server running or quit abnormally, part of data may or may not been synced to downstream
+	StatusRunning int = 1
+)
+
 var (
 	// ErrCheckPointClosed indicates the CheckPoint already closed.
 	ErrCheckPointClosed = errors.New("CheckPoint already closed")
@@ -31,10 +39,13 @@ type CheckPoint interface {
 	Load() error
 
 	// Save saves checkpoint information.
-	Save(int64, int64) error
+	Save(commitTS int64, slaveTS int64, status int) error
 
-	// Pos gets position information.
+	// TS gets checkpoint commit timestamp.
 	TS() int64
+
+	// Status return the status saved.
+	Status() int
 
 	// Close closes the CheckPoint and release resources, after closed other methods should not be called again.
 	Close() error
--- a/drainer/checkpoint/mysql.go
+++ b/drainer/checkpoint/mysql.go
@@ -6,9 +6,7 @@
 //
 //     http://www.apache.org/licenses/LICENSE-2.0
 //
-// Unless required by applicable law or agreed to in writing, software
-// distributed under the License is distributed on an "AS IS" BASIS,
-// See the License for the specific language governing permissions and
+// Unless required by applicable law or agreed to in writing, software // distributed under the License is distributed on an "AS IS" BASIS, // See the License for the specific language governing permissions and
 // limitations under the License.
 
 package checkpoint
@@ -35,8 +33,9 @@ type MysqlCheckPoint struct {
 	schema string
 	table  string
 
-	CommitTS int64            `toml:"commitTS" json:"commitTS"`
-	TsMap    map[string]int64 `toml:"ts-map" json:"ts-map"`
+	StatusSaved int              `toml:"status" json:"status"`
+	CommitTS    int64            `toml:"commitTS" json:"commitTS"`
+	TsMap       map[string]int64 `toml:"ts-map" json:"ts-map"`
 }
 
 var sqlOpenDB = pkgsql.OpenDB
@@ -106,7 +105,7 @@ func (sp *MysqlCheckPoint) Load() error {
 }
 
 // Save implements checkpoint.Save interface
-func (sp *MysqlCheckPoint) Save(ts, slaveTS int64) error {
+func (sp *MysqlCheckPoint) Save(ts, slaveTS int64, status int) error {
 	sp.Lock()
 	defer sp.Unlock()
 
@@ -115,6 +114,7 @@ func (sp *MysqlCheckPoint) Save(ts, slaveTS int64) error {
 	}
 
 	sp.CommitTS = ts
+	sp.StatusSaved = status
 
 	if slaveTS > 0 {
 		sp.TsMap["master-ts"] = ts
@@ -133,6 +133,14 @@ func (sp *MysqlCheckPoint) Save(ts, slaveTS int64) error {
 	}
 
 	return nil
+}
+
+// Status implements CheckPoint.Status interface
+func (sp *MysqlCheckPoint) Status() int {
+	sp.RLock()
+	defer sp.RUnlock()
+
+	return sp.StatusSaved
 }
 
 // TS implements CheckPoint.TS interface
--- a/drainer/checkpoint/mysql.go
+++ b/drainer/checkpoint/mysql.go
@@ -6,7 +6,9 @@
 //
 //     http://www.apache.org/licenses/LICENSE-2.0
 //
-// Unless required by applicable law or agreed to in writing, software // distributed under the License is distributed on an "AS IS" BASIS, // See the License for the specific language governing permissions and
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// See the License for the specific language governing permissions and
 // limitations under the License.
 
 package checkpoint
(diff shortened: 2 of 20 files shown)


suzaku:
@july2993 Please resolve conflicts.

july2993:
> @july2993 Please resolve conflicts.

resolved

drainer/relay.go
```diff
@@ -0,0 +1,150 @@
...
+			loaderClosed = true
+		}
+
+		// break once we drainer the success items return by loader.
+		if loaderClosed && successTxnC == nil {
+			break
+		}
+
+		select {
+		case sbinlog, ok := <-readerTxnsC:
```
suzaku: Is it possible for two `sbinlog` to be received successively in this loop?

drainer/relay_test.go
```diff
@@ -0,0 +1,109 @@
...
+	"github.com/pingcap/tidb-binlog/drainer/translator"
+	"github.com/pingcap/tidb-binlog/pkg/binlogfile"
+	"github.com/pingcap/tidb-binlog/pkg/loader"
+)
+
+type relaySuite struct{}
+
+var _ = check.Suite(&relaySuite{})
+
+type noOpLoader struct {
```
suzaku: Embed the interface so that we don't have to repeat methods we don't need in this test?

drainer/checkpoint/checkpoint.go
```diff
@@ -19,6 +19,14 @@ import (
 	"go.uber.org/zap"
 )
 
+const (
+	// StatusNormal means server quit normally, data <= ts is synced to downstream
+	StatusNormal int = 0
```
GregoryIan: Worth a better name, like `StatusConsistent` and `StatusInconsistent`

GregoryIan: Good Change!

drainer/checkpoint/mysql.go
```diff
@@ -68,6 +74,16 @@ func newMysql(cfg *Config) (CheckPoint, error) {
 		return nil, errors.Annotatef(err, "exec failed, sql: %s", sql)
 	}
 
+	if sp.clusterID == 0 {
+		id, err := getClusterID(db, sp.schema, sp.table)
```
GregoryIan: This logic is very weird. 
In order to handle the situation where the upstream cluster is completely suspended, drainer needs a correct way to fetch the  cluster id. But current implementation is not satisfying, like drainer can't allow store multiple entries in checkpoint table

GregoryIan: I file an issue about it https://github.com/pingcap/tidb-binlog/issues/889

drainer/checkpoint/util.go
```diff
@@ -14,12 +14,18 @@
...
 	"fmt"
 
 	// mysql driver
 	_ "github.com/go-sql-driver/mysql"
+	"github.com/pingcap/errors"
 )
 
+// ErrNoCheckpointItem represents there's no any checkpoint item and the cluster id must be specified
+// for the mysql checkpoint type.
+var ErrNoCheckpointItem = errors.New("no any checkpoint item")
```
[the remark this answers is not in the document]
GregoryIan: Is there any special reason @suzaku ?  Looking forward to learning something from your comment 😄

drainer/checkpoint/util_test.go
```diff
@@ -0,0 +1,63 @@
...
+	tests := []struct {
+		name              string
+		rows              []uint64
+		id                uint64
+		err               bool
+		checkSpecifiedErr error
+	}{
+		{"no row", nil, 0, true, ErrNoCheckpointItem},
+		{"on row", []uint64{1}, 1, false, nil},
+		{"multi row", []uint64{1, 2}, 0, true, nil},
```
GregoryIan: should the `checkSpecifiedErr` be nil?

drainer/config.go
```diff
@@ -139,9 +148,6 @@ func NewConfig() *Config {
 	fs.StringVar(&cfg.SyncerCfg.IgnoreSchemas, "ignore-schemas", "INFORMATION_SCHEMA,PERFORMANCE_SCHEMA,mysql", "disable sync those schemas")
 	fs.IntVar(&cfg.SyncerCfg.WorkerCount, "c", 16, "parallel worker count")
 	fs.StringVar(&cfg.SyncerCfg.DestDBType, "dest-db-type", "mysql", "target db type: mysql or tidb or file or kafka; see syncer section in conf/drainer.toml")
-	fs.StringVar(&cfg.SyncerCfg.RelayLogDir, "relay-log-dir", "", "path to relay log of syncer")
```
GregoryIan: can we leave them? 

they may improve ease of use, like in K8s, because configuration files are always cumbersome

pkg/util/util.go
```diff
@@ -251,20 +252,12 @@ func GetPdClient(etcdURLs string, securityConfig security.Config) (pd.Client, er
 		return nil, errors.Trace(err)
 	}
 
-	pdReconnTimes := 30
-
 	var pdCli pd.Client
-	for i := 1; i < pdReconnTimes; i++ {
```
GregoryIan: Removing retries may reduce usability, as long as drainer can't connect to pd, it will try the relay log and then exit.

GregoryIan: we can leave the simple retry logic

drainer/checkpoint/util.go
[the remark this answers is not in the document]
suzaku: @GregoryIan When returning an error created using the `pingcap/errors` package, the stack trace would end in the line where the error is declared.

drainer/checkpoint/util_test.go
july2993: yes,  true, nil means the result will be some error but we don't check what kind of error.

pkg/util/util.go
july2993: As the commen added, the pd.NewClient will retry 100 time already.

drainer/relay.go
```diff
@@ -0,0 +1,150 @@
...
+	if cp.Status() == checkpoint.StatusNormal {
+		return nil
+	}
+
+	reader, err := relay.NewReader(scfg.Relay.LogDir, 1 /* readBufferSize */)
+	if err != nil {
+		return errors.Annotate(err, "failed to create reader")
+	}
+
+	db, ld, err := sync.CreateLoader(scfg.To, scfg.WorkerCount, scfg.TxnBatch,
```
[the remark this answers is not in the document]
july2993: ? should check err first, then defer

drainer/config.go
```diff
@@ -74,14 +74,23 @@ type SyncerConfig struct {
...
 }
 
+// RelayConfig is the Relay log's configuration.
+type RelayConfig struct {
+	LogDir      string `toml:"log-dir" json:"log-dir"`
+	MaxFileSize int64  `toml:"max-file-size" json:"max-file-size"`
+}
+
+// SwitchOn return true if we need to handle relay log.
+func (rc RelayConfig) SwitchOn() bool {
```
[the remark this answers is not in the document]
july2993: changed

drainer/config.go
july2993: reverted

drainer/relay.go
july2993: fixed

drainer/checkpoint/util.go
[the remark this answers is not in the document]
july2993: use std in https://github.com/pingcap/tidb-binlog/pull/883/commits/d34142569b67bb097232bd91c791dce05c665f38

drainer/relay.go
```diff
@@ -0,0 +1,156 @@
...
+			ld.Close()
+			loaderClosed = true
+		}
+
+		// break once we drainer the success items return by loader.
+		if loaderClosed && successTxnC == nil {
+			break
+		}
+
+		select {
```
suzaku: The logic in this `select` is a bit hard to follow and reason about. Will it be simpler if we just we two separate goroutine for reading and writing?

drainer/relay_test.go
july2993: all need unless `SaveMode`, I will keep this.
instead, we should make the pkg/Loader easy to test(not make it an interface only for test), we can use the struct directly and add some method like set it silently accept `input` and return success directly.

drainer/sync/mysql.go
```diff
@@ -41,11 +41,19 @@ type MysqlSyncer struct {
...
+// CreateLoader create the Loader instance.
+func CreateLoader(
+	cfg *DBConfig,
+	worker int,
+	batchSize int,
+	queryHistogramVec *prometheus.HistogramVec,
+	sqlMode *string,
+	destDBType string,
+	info *loopbacksync.LoopBackSync,
+) (db *sql.DB, ld loader.Loader, err error) {
```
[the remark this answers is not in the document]
july2993: update in https://github.com/pingcap/tidb-binlog/pull/883/commits/bb81853698b4d86fb8d81481b49b48d31e674c68

drainer/relay.go
```diff
@@ -0,0 +1,150 @@
...
+	checkpointTS := cp.TS()
+	lastSuccessTS := checkpointTS
+	r.Run()
+
+	loaderQuit := make(chan struct{})
+	var loaderErr error
+	go func() {
+		ld.SetSafeMode(true)
+		loaderErr = ld.Run()
+		close(loaderQuit)
```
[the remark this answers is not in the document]
july2993:
```
120         case success, ok := <-successTxnC:
121             if !ok {
122                 successTxnC = nil
123                 log.Info("success closed")
124                 continue
125             }
126             lastSuccessTS = success.Metadata.(int64)
``` 
will continue the loop and run into the `loaderQuit` case at nest select

drainer/relay.go
```diff
@@ -0,0 +1,150 @@
...
+		case <-loaderQuit:
+			if loaderErr != nil {
+				return errors.Trace(loaderErr)
+			}
+		}
+	}
+
+	log.Info("finish feed by relay log")
+
+	readerErr := <-r.Error()
```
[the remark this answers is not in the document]
july2993: `feedByRelayLog` will need to close `ld` once having put all txns into it, and  `r.Error()` should better put here together with the according `r.Run()` and `r.Txns()`

drainer/checkpoint/checkpoint.go
july2993: rename as `StatusConsistent` https://github.com/pingcap/tidb-binlog/pull/883/commits/002d4493a4f5cfcb06bd41acc762788271b5c2d9

drainer/relay.go
july2993: actually first try in multi goroutine style first, but finally, give up, will keep the code style now.

pkg/util/util.go
GregoryIan: now it's reasonable

PR #884, opened 2020-01-15
binlog loop back sync

july2993:
cherry pick #879 #867 

cherry pick #879 #867 

    Add sync mode config (#867)

    allow when the column number of downstream table mismatch with current schema.

    For the case bidirectional replication, we will execute the DDL at one side, for add or drop column
    the column number will mismatch.

    cluster A <-> cluster B

    drop column of table t at cluster A
    some DML of table t at cluster B will miss the column dropped compared to cluster A

    binlog loop back sync (#879)

    In the AA dual activity scenario, there is data write (DML) on both sides. The data of tidb cluster on both sides needs to be synchronized with each other，but avoid loopback synchronization，and ddl is only writed one side.

    add three variables in drainer.toml as identification to confirm need sync ddl or not and need set sync mark identification or not and sync identification id to Avoid loopback synchronization
    add configuration item
    loopback-control (true/false) set mark table identification or not and filter txn by mark table
    ddl-sync (true/false) sync ddl to downstream DB or not
    channel-id (integer) sync identification id，avoid loopback synchronization

### Check List 

Code changes

### DIFF
--- a/cmd/drainer/drainer.toml
+++ b/cmd/drainer/drainer.toml
@@ -84,6 +84,10 @@ password = ""
 # if encrypted_password is not empty, password will be ignored.
 encrypted_password = ""
 port = 3306
+# 1: SyncFullColumn, 2: SyncPartialColumn
+# when setting SyncPartialColumn drainer will allow the downstream schema
+# having more or less column numbers and relax sql mode by removing STRICT_TRANS_TABLES.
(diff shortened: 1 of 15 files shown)


july2993:
@suzaku @freemindLi PTAL

freemindLi:
LGTM

PR #885, opened 2020-01-15
drainer: Fix #724, Enable drainer to purge old incremental backup data on disk

### LINKED ISSUE #724: Enable drainer to purge old incremental backup data on disk
## Feature Request

**Is your feature request related to a problem? Please describe:**

Currently, incremental backup data needs to be purged manually if drainer is configured to output files to disk.

**Describe the feature you'd like:**

Implement a retention period feature where drainer automatically purges data that is older than a configured time.
e.g. If 7 days is configured as a retention period, incremental backup data that is older than 7 days is automatically deleted to free up disk space.

**Describe alternatives you've considered:**

N/A

**Teachability, Documentation, Adoption, Migration Strategy:**

For example, a new configuration like the following can be added:
```
# Determines when to remove old data. Effective only if db-type is "file". Defaults to 7d.
retention-time = 30d

### PR DESCRIPTION
**Pull request by suzaku**

Enable drainer to purge old incremental backup data on disk.

1. Add a `retention-time` config to specify how long binlog files should be kept in drainer. Backward-compatibility is maintained by treating non-positive values of `retention-time` as *no GC*.
1. Currently `retention-time` is of type `int`, which is consistent with the pump `GC` configuration. We may later introduce `p8s` style time format for both `GC` and `retention-time` and use days as the default unit to maintain backward-compatibility.

### Check List 

Tests 

 - Unit test

Code changes

      The `GC` method is changed to two new methods `GCByTIme` and `GCByPos`.

Related changes

### DIFF
--- a/cmd/drainer/drainer.toml
+++ b/cmd/drainer/drainer.toml
@@ -129,6 +129,8 @@ port = 3306
 #[syncer.to]
 # directory to save binlog file, default same as data-dir(save checkpoint file) if this is not configured.
 # dir = "data.drainer"
+#
+# retention-time = 7
 
 
 # when db-type is kafka, you can uncomment this to config the down stream kafka, it will be the globle config kafka default
@@ -138,8 +140,6 @@ port = 3306
 # kafka-addrs = "127.0.0.1:9092"
 # kafka-version = "0.8.2.0"
 # kafka-max-messages = 1024
-#
-# retention-time = 7
 #
 # the topic name drainer will push msg, the default name is <cluster-id>_obinlog
 # be careful don't use the same name if run multi drainer instances
--- a/cmd/drainer/drainer.toml
+++ b/cmd/drainer/drainer.toml
@@ -139,6 +139,7 @@ port = 3306
 # kafka-version = "0.8.2.0"
 # kafka-max-messages = 1024
 #
+# retention-time = 7
 #
 # the topic name drainer will push msg, the default name is <cluster-id>_obinlog
 # be careful don't use the same name if run multi drainer instances
--- a/drainer/sync/pb.go
+++ b/drainer/sync/pb.go
@@ -14,7 +14,11 @@
 package sync
 
 import (
+	"context"
+	"time"
+
 	"github.com/pingcap/errors"
+	"github.com/pingcap/log"
 	"github.com/pingcap/tidb-binlog/drainer/translator"
 	"github.com/pingcap/tidb-binlog/pkg/binlogfile"
 	pb "github.com/pingcap/tidb-binlog/proto/binlog"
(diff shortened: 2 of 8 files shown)

### CODE REVIEW

**Comment by suzaku** [comment]
@shinnosuke-okada @july2993 PTAL

File: cmd/drainer/drainer.toml [hunk 1]
```diff
@@ -139,6 +139,7 @@ port = 3306
 # kafka-version = "0.8.2.0"
 # kafka-max-messages = 1024
 #
+# retention-time = 7
```
july2993: put it at the place of the #[syncer.to] example part for type = file

drainer/sync/pb.go
```diff
@@ -24,21 +28,44 @@ import (
...
 	*baseSyncer
+
+	binlogger binlogfile.Binlogger
+	cancel    func()
 }
 
 // NewPBSyncer sync binlog to files
-func NewPBSyncer(dir string, tableInfoGetter translator.TableInfoGetter) (*pbSyncer, error) {
-	binlogger, err := binlogfile.OpenBinlogger(dir, binlogfile.SegmentSizeBytes)
+func NewPBSyncer(cfg *DBConfig, tableInfoGetter translator.TableInfoGetter) (Syncer, error) {
```
july2993: - return the type instead of the interface
- pass the specified param you need

suzaku: Would you mind explaining why?

july2993: - return the type instead of the interface
it will be flexible to use pBSyncer directly if need? 
- pass the specified param you need
decouple it from `DBConfig` and the redundant `DBConfig` contains too many fields unrelated to PBSyncer, user will never know how to use it actually unless reading the code of `NewPBSyncer`.

july2993: > * return the type instead of the interface
>   it will be flexible to use pBSyncer directly if need?
https://github.com/golang/go/wiki/CodeReviewComments#interfaces
Go interfaces generally belong in the package that uses values of the interface type, not the package that implements those values. The implementing package should return concrete (usually pointer or struct) types: that way, new methods can be added to implementations without requiring extensive refactoring.
> * pass the specified param you need
>   decouple it from `DBConfig` and the redundant `DBConfig` contains too many fields unrelated to PBSyncer, user will never know how to use it actually unless reading the code of `NewPBSyncer`.

suzaku:
@july2993 PTAL

pkg/binlogfile/binlogger.go
```diff
@@ -348,8 +351,37 @@ func (b *binlogger) Walk(ctx context.Context, from binlog.Pos, sendBinlog func(e
...
+	// skip the latest binlog file
+	for _, name := range names[:len(names)-1] {
+		curSuffix, _, err := ParseBinlogName(name)
+		if err != nil {
+			log.Error("parse binlog failed", zap.Error(err))
+		}
+		if curSuffix < pos.Suffix {
+			fileName := path.Join(b.dir, name)
+			if err := os.Remove(fileName); err != nil {
+				log.Error("remove old binlog file err", zap.Error(err), zap.String("file name", fileName))
```
GregoryIan:
```suggestion
				log.Error("fail to remove old binlog file ", zap.Error(err), zap.String("file name", fileName))
```

GregoryIan: Better not use abbreviations, like `err` 😄

pkg/binlogfile/binlogger.go
```diff
@@ -369,22 +401,12 @@ func (b *binlogger) GC(days time.Duration, pos binlog.Pos) {
...
-			log.Error("parse binlog failed", zap.Error(err))
-		}
-
-		if curSuffix < pos.Suffix {
-			err := os.Remove(fileName)
-			if err != nil {
-				log.Error("remove old binlog file err")
+		if time.Since(fi.ModTime()) > retentionTime {
+			if err := os.Remove(fileName); err != nil {
+				log.Error("remove old binlog file err", zap.Error(err), zap.String("file name", fileName))
```
GregoryIan:
```suggestion
				log.Error("fail to remove old binlog file", zap.Error(err), zap.String("file name", fileName))
```

GregoryIan:
Rest LGTM

PR #890, opened 2020-02-05
Fix current DML&DDL corner case for bidirectional replication 

july2993:
In bidirectional replication
cluster A <-> cluster B
only do and replicate DDL in A

- in B will filter all DDL and will never refresh the table info in pkg/Loader
- we will not know the schema is changed while replicating data from B to A if a DDL is applied on A and not having replicated to B. A sample case:
    * drop a column `c81` on table `auto1` (apply on A)
    * apply some DML on B and replicate to cluster A will failed for we specified the column `c81` to insert cause we have not learned that the Shema is changed on A.
   
- refresh table info even if we filter the ddl on B(make an empty query force loader to refresh now)
- try refresh table info if failed to exec the dml for some error (column mismatch now)
- always  use the values pass into Loader to construct the SQL(compared to the downstream table info we will specified a `NULL` value as the column value if we have not specified the value in `Values` pass into `Loader`) an example before:
   * A   auto(a int primary key, c int default 10)
   * B  auto(a int primary key)
   * while replicate data from B to A we will   `insert into auto(a,c) (11, NULL)` (set a NULL for the column c)
  * the column c will not consistent between A and B even later apply the add column on B

### Check List 

Tests 

 - Unit test
use tool https://github.com/july2993/bitest

Related changes

port back to master

### DIFF
--- a/drainer/syncer.go
+++ b/drainer/syncer.go
@@ -390,11 +390,6 @@ ForLoop:
 			// DDL (with version 10, commit ts 100) -> DDL (with version 9, commit ts 101) would never happen
 			s.schema.addJob(b.job)
 
-			if !s.cfg.SyncDDL {
-				log.Info("Syncer skips DDL", zap.String("sql", b.job.Query), zap.Int64("ts", b.GetCommitTs()), zap.Bool("SyncDDL", s.cfg.SyncDDL))
-				continue
-			}
-
 			log.Debug("get DDL", zap.Int64("SchemaVersion", b.job.BinlogInfo.SchemaVersion))
 			lastDDLSchemaVersion = b.job.BinlogInfo.SchemaVersion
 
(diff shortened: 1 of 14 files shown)


drainer/syncer.go
```diff
@@ -418,23 +413,36 @@ ForLoop:
...
-				err = s.dsyncer.Sync(&dsync.Item{Binlog: binlog, PrewriteValue: nil, Schema: schema, Table: table})
-				if err != nil {
-					err = errors.Annotatef(err, "add to dsyncer, commit ts %d", binlog.CommitTs)
-					break ForLoop
+			if !s.cfg.SyncDDL {
+				log.Info("skip ddl by SyncDDL setting to false", zap.String("schema", schema), zap.String("table", table),
+					zap.String("sql", sql), zap.Int64("commit ts", commitTS))
+				// A empty sql force it to evict the downstream table info.
+				if s.cfg.DestDBType == "tidb" || s.cfg.DestDBType == "mysql" {
+					binlog.DdlQuery = []byte("")
```
GregoryIan: how about adding a filed `ShouldSkiped` into `dsync.Item`, and ingnore these `dsync.Item` while we need to execute it?  code implementation may be simpler.

GregoryIan:
Rest LGTM

pkg/loader/executor.go
```diff
@@ -269,10 +277,54 @@ func (e *executor) splitExecDML(ctx context.Context, dmls []*DML, exec func(dmls
...
+
+			if tryRefreshTableErr(execErr) && e.refreshTableInfo != nil {
+				log.Info("try refresh table info")
+				name2info := make(map[string]*tableInfo)
+				for _, dml := range dmls {
+					name := dml.TableName()
+					info, ok := name2info[name]
+					if !ok {
+						var err error
+						info, err = e.refreshTableInfo(dml.Database, dml.Table)
```
GregoryIan: There may be problems calling `refreshTableInfo` many times in a short time, If so will it have a big impact?

GregoryIan:
👍  https://github.com/july2993/bitest

drainer/syncer.go
july2993: also need some king like SHouldSkiped into pkg/Loader instead of empty string?

pkg/loader/executor.go
july2993: as discuss offline I don't think so.

drainer/syncer.go
july2993: https://github.com/pingcap/tidb-binlog/pull/890/commits/a17b05a2f99508e42c7038759398245ede7e59d4
Use explictly field to skip item @GregoryIan @suzaku PTAL

july2993:
@GregoryIan PTAL

pkg/loader/executor.go
GregoryIan: Correct it, the conclusion of our offline discussion is that need some
 testing for it

PR #892, opened 2020-02-07
Fix current DML&DDL corner case for bidirectional replication  (#890)

july2993:
pick #890 

pick #890 

### Check List 

Related changes

### DIFF
--- a/pkg/loader/load.go
+++ b/pkg/loader/load.go
@@ -273,10 +273,6 @@ func (s *loaderImpl) Close() {
 
 var utilGetTableInfo = getTableInfo
 
-func (s *loaderImpl) invalidateTableInfo(schema string, table string) {
-	s.tableInfos.Delete(quoteSchema(schema, table))
-}
-
 func (s *loaderImpl) refreshTableInfo(schema string, table string) (info *tableInfo, err error) {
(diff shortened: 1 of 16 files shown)


pkg/loader/load.go
```diff
@@ -300,6 +302,10 @@ func (s *loaderImpl) refreshTableInfo(schema string, table string) (info *tableI
 	return
 }
 
+func (s *loaderImpl) evitTableInfo(schema string, table string) {
```
suzaku: In `master`, we already have a method called `invalidateTableInfo` for invalidating the cached table info.

july2993: [remove invalidateTableInfo
](https://github.com/pingcap/tidb-binlog/pull/892/commits/7aa677b6cce31bf0d5439b94e219c52dda5e5ed7)

suzaku: I suppose we should remove `evitTableInfo` instead, `evit` is not even an English word ...

july2993: https://github.com/pingcap/tidb-binlog/pull/892/commits/686ef25a34a4c242fd8e29bd1ffb3ab89bcd93e4
use `evict`

july2993:
@suzaku @GregoryIan PTAL

PR #895, opened 2020-02-12
Rename Status as Consistent

july2993:
see https://github.com/pingcap/docs-cn/pull/2224#discussion_r376911866

Rename Status as Consistent

### Check List 

Tests 

 - Unit test

### DIFF
--- a/drainer/checkpoint/checkpoint.go
+++ b/drainer/checkpoint/checkpoint.go
@@ -19,14 +19,6 @@ import (
 	"go.uber.org/zap"
 )
 
-const (
-	// StatusConsistent means server quit normally, data <= ts is synced to downstream
-	StatusConsistent int = 0
-
-	// StatusRunning means server running or quit abnormally, part of data may or may not been synced to downstream
-	StatusRunning int = 1
-)
-
 var (
 	// ErrCheckPointClosed indicates the CheckPoint already closed.
 	ErrCheckPointClosed = errors.New("CheckPoint already closed")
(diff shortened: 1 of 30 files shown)


drainer/checkpoint/checkpoint.go
```diff
@@ -19,14 +19,6 @@ import (
 	"go.uber.org/zap"
 )
 
-const (
-	// StatusConsistent means server quit normally, data <= ts is synced to downstream
-	StatusConsistent int = 0
```
GregoryIan: `Status` is more flexible and scalable. I think It's better if we map status to readable words.

july2993: what about keep the origin Status and add a filed `StatusMessage` in the checkpoint ?

drainer/checkpoint/checkpoint.go
```diff
@@ -39,13 +31,13 @@ type CheckPoint interface {
...
-	Save(commitTS int64, slaveTS int64, status int) error
+	Save(commitTS int64, slaveTS int64, consistent bool) error
 
 	// TS gets checkpoint commit timestamp.
 	TS() int64
 
-	// Status return the status saved.
-	Status() int
+	// Consistent return the Consistent status saved.
+	Consistent() bool
```
suzaku: `IsConsistent`?

GregoryIan:
LGTM

drainer/checkpoint/checkpoint.go
july2993: change  by https://github.com/pingcap/tidb-binlog/pull/895/commits/d0d404718584604fe5428d7d5c276129f9c36dea

PR #903, opened 2020-02-17
Optimize the performance of bidirectional synchronous update table of _drainer_repl_mark

freemindLi:
…of _drainer_repl_mark

1.Optimize the performance of bidirectional synchronous update table of _drainer_repl_mark

The number of channel related records in the mark table is initialized according to the number of worker "count. During synchronization, each syncer operates one piece of data, which will not cause multithreading to update one record, resulting in synchronization performance bottleneck

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

### DIFF
--- a/drainer/loopbacksync/loopbacksync.go
+++ b/drainer/loopbacksync/loopbacksync.go
@@ -16,7 +16,7 @@ package loopbacksync
 const (
 	//MarkTableName mark table name
 	MarkTableName = "retl._drainer_repl_mark"
-	//ID syncer worker coroutine id
+	//CoroutineID syncer worker coroutine id
 	CoroutineID = "coroutine_id"
 	//ChannelID channel id
 	ChannelID = "channel_id"
--- a/drainer/loopbacksync/loopbacksync.go
+++ b/drainer/loopbacksync/loopbacksync.go
@@ -16,8 +16,8 @@ package loopbacksync
 const (
 	//MarkTableName mark table name
 	MarkTableName = "retl._drainer_repl_mark"
-	//CoroutineID syncer worker coroutine id
-	CoroutineID = "coroutine_id"
+	//ID syncer worker coroutine id
+	ID = "id"
 	//ChannelID channel id
 	ChannelID = "channel_id"
 	//Val val
--- a/drainer/loopbacksync/loopbacksync.go
+++ b/drainer/loopbacksync/loopbacksync.go
@@ -16,6 +16,8 @@ package loopbacksync
 const (
 	//MarkTableName mark table name
 	MarkTableName = "retl._drainer_repl_mark"
+	//ID syncer worker thread id
+	ID = "id"
 	//ChannelID channel id
 	ChannelID = "channel_id"
 	//Val val
--- a/drainer/loopbacksync/loopbacksync.go
+++ b/drainer/loopbacksync/loopbacksync.go
@@ -17,7 +17,7 @@ const (
 	//MarkTableName mark table name
 	MarkTableName = "retl._drainer_repl_mark"
 	//ID syncer worker coroutine id
-	ID = "coroutineID"
+	CoroutineID = "coroutine_id"
 	//ChannelID channel id
 	ChannelID = "channel_id"
 	//Val val
--- a/drainer/loopbacksync/loopbacksync.go
+++ b/drainer/loopbacksync/loopbacksync.go
@@ -16,8 +16,8 @@ package loopbacksync
 const (
 	//MarkTableName mark table name
 	MarkTableName = "retl._drainer_repl_mark"
-	//ID syncer worker thread id
-	ID = "id"
+	//ID syncer worker coroutine id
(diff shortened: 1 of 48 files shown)


drainer/loopbacksync/loopbacksync.go
```diff
@@ -16,6 +16,8 @@ package loopbacksync
 const (
 	//MarkTableName mark table name
 	MarkTableName = "retl._drainer_repl_mark"
+	//ID syncer worker thread id
+	ID = "id"
```
WangXiangUSTC: how about change to a meaningful name like `threadID`?

pkg/loader/executor.go
```diff
@@ -33,19 +34,23 @@ import (
 )
 
 var defaultBatchSize = 128
+var defaultWorkerCount = 16
+var index int64
```
WangXiangUSTC: how about using 
```
var (
    ...
)
```

pkg/loader/executor.go
```diff
@@ -118,19 +127,72 @@ func (tx *tx) commit() error {
 }
 
 func (e *executor) updateMark(channel string, tx *tx) error {
+	if e.info == nil {
+		return nil
+	}
+	index = e.getIndex(index)
```
WangXiangUSTC: may have more than one thread use `index`, how about using `atomic.StoreInt64` and `atomic.LoadInt64` to set or get `index`?

drainer/loopbacksync/loopbacksync.go
freemindLi: This is not a real thread ID, just a 0-workercount-1 sequence for go's concurrent programs；Maybe it's called coroutineID is appropriate？

WangXiangUSTC:
LGTM, @july2993 PTAL

pkg/loader/executor.go
```diff
@@ -118,19 +129,73 @@ func (tx *tx) commit() error {
...
+	sql := fmt.Sprintf("delete from %s where %s=? ", loopbacksync.MarkTableName, loopbacksync.ChannelID)
+	args = append(args, e.info.ChannelID)
+	_, err1 := tx.autoRollbackExec(sql, args...)
+	if err1 != nil {
+		return errors.Trace(err1)
+	}
+	err2 := tx.commit()
+	return errors.Trace(err2)
+}
+func (e *executor) addIndex() int64 {
```
july2993: why not just return `atomic.AddInt64(&index, 1) % e.workerCount`.

drainer/loopbacksync/loopbacksync.go
```diff
@@ -16,6 +16,8 @@ package loopbacksync
 const (
 	//MarkTableName mark table name
 	MarkTableName = "retl._drainer_repl_mark"
+	//CoroutineID syncer worker coroutine id
+	CoroutineID = "coroutine_id"
```
july2993: just `id` ?

pkg/loader/executor.go
freemindLi: addIndex function？just return atomic.AddInt64(&index, 1) % e.workerCount need assigne to index variable，Multithreading maybe one is setting index，another thread is readding，Data confusion

pkg/loader/executor.go
```diff
@@ -118,19 +129,71 @@ func (tx *tx) commit() error {
 }
 
 func (e *executor) updateMark(channel string, tx *tx) error {
+	if e.info == nil {
+		return nil
+	}
+	var args []interface{}
+	sql := fmt.Sprintf("update %s set %s=%s+1 where %s=? and %s=? limit 1;", loopbacksync.MarkTableName, loopbacksync.Val, loopbacksync.Val, loopbacksync.ID, loopbacksync.ChannelID)
+	args = append(args, e.addIndex(), e.info.ChannelID)
+	_, err1 := tx.autoRollbackExec(sql, args...)
```
july2993: maybe we can check the affect rows to make sure we do update some row?

pkg/loader/executor.go
```diff
@@ -118,19 +129,71 @@ func (tx *tx) commit() error {
...
+		builder.WriteString(holder)
+	}
 	var args []interface{}
-	sql := fmt.Sprintf("INSERT INTO %s%s on duplicate key update %s=%s+1;", loopbacksync.MarkTableName, columns, loopbacksync.Val, loopbacksync.Val)
-	args = append(args, e.info.ChannelID, status, channel)
-	_, err := tx.autoRollbackExec(sql, args...)
+	for id := 0; id < e.workerCount; id++ {
+		args = append(args, id, e.info.ChannelID, status, channel)
+	}
+	tx, err := e.begin()
```
july2993: in e.begin() it will `updateMark ` it's confusing doing this.
suggest writing a func to accepte  a `*sql.DB` directly to initMarkTable

PR #911, opened 2020-02-25
Plugin

tsthght:
### What problem does this PR solve? 
Drainer support plugin

### What is changed and how it works?
Add plugin framework to drainer, user can customize plugins, and use parameter `plugin-name` , `plugin-path` and `support-plugin` to load plugin into drainer.

### Check List 

Tests 

 - Unit test
 - Integration test
 - Manual test (add detailed scripts or steps below)
 - No code

Code changes

 - Has exported function/method change
 - Has exported variable/fields change
 - Has interface methods change
 - Has persistent data change

Side effects

 - Possible performance regression
 - Increased code complexity
 - Breaking backward compatibility

Related changes

 - Need to cherry-pick to the release branch
 - Need to update the documentation
 - Need to update the `tidb-ansible` repository
 - Need to be included in the release note

PR #916, opened 2020-02-28
plugin for loopbacksync

cvvz:

There is no plugin for loopbacksync.

Add a plugin for loopbacksync.
### Check List 

Tests
 - Unit test


WangXiangUSTC:
thanks, we will review this pr after https://github.com/pingcap/tidb-binlog/pull/911 is merged @cvvz

PR #920, opened 2020-03-03
Update parser add test for auto_random

july2993:
Update parser add test for auto_random

Update parser add test for auto_random

### Check List 

Tests 

 - Integration test
 

Code changes

Side effects


WangXiangUSTC:
LGTM

PR #921, opened 2020-03-03
Update parser/tidb add test for auto_random

july2993:
Update parser/tidb add test for auto_random

Update parser/tidb add test for auto_random

### Check List 

Tests 

 - Integration test

Code changes

Related changes


july2993:
ci fail because sync diff failed to check bu it can actually replicate normally.

table diff return error
https://github.com/pingcap/tidb-binlog/blob/649ec60f9ceb65fbbaf22e06e43e39ae62e56df9/tests/util/db.go#L115
```
  2 [2020/03/04 17:37:35] [error] db.go:116 failed to check equal{error 25 0  [ddl:8216]Invalid auto random: auto_random is an experimental feature, which can only be used when allow-auto-random is enabled. This can be changed in the configuration.} {sta
```

but not figuring out why syn_diff failed to check yes.

july2993:
> ci fail because sync diff failed to check bu it can actually replicate normally.
> 
> table diff return error
> https://github.com/pingcap/tidb-binlog/blob/649ec60f9ceb65fbbaf22e06e43e39ae62e56df9/tests/util/db.go#L115
> 
> ```
>   2 [2020/03/04 17:37:35] [error] db.go:116 failed to check equal{error 25 0  [ddl:8216]Invalid auto random: auto_random is an experimental feature, which can only be used when allow-auto-random is enabled. This can be changed in the configuration.} {sta
> ```
> 
> but not figuring out why syn_diff failed to check yes.

ref https://github.com/pingcap/tidb-tools/pull/321

Issue #923, opened 2020-03-04
binlogctl failed to parse pd-urls


july2993:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

1. What did you do?
```
./binlogctl -pd-urls="http://12.12.12.12:2379" -cmd pumps
```
2. What did you expect to see?
work normally

3. What did you see instead?
```
2020/03/04 17:57:31 main.go:39: [error] parse cmd flags, error: parse EtcdURLs error: http://168.63.117.150:2379, address http://168.63.117.150:2379: too many colons in address
```

4. Please provide the relate downstream type and version of drainer.
(run `drainer -V` in terminal to get drainer's version)
2.1.19

kennytm:
why the input is `12.12.12.12` but the error message is `168.63.117.150`

PR #924, opened 2020-03-04
fix parse pd-urls in binlogctl

july2993:
fix https://github.com/pingcap/tidb-binlog/issues/923
fail since #750 
just copy code from master in `pkg/flag`

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

Side effects

Related changes


pkg/flags/urls.go
```diff
@@ -64,17 +77,35 @@ func NewURLsValue(init string) (*URLsValue, error) {
...
-			return nil, errors.Trace(err)
+			return nil, errors.Errorf("parse url %s failed %v", str, err)
 		}
-		addrs = append(addrs, str)
+		if u.Scheme != "http" && u.Scheme != "https" && u.Scheme != "unix" && u.Scheme != "unixs" {
+			return nil, errors.Errorf("URL scheme must be http, https, unix, or unixs: %s", str)
+		}
+		if _, _, err := net.SplitHostPort(u.Host); err != nil {
+			return nil, errors.Errorf(`URL address does not have the form "host:port": %s`, str)
+		}
```
kennytm: unix??

Even if you're going to support unix-domain sockets, the `unix:///tmp/etcd.sock` will have an empty `u.Host` and cause `net.SplitHostPort` to fail ("missing port in address").

july2993: I don'ts what it wants actually, but after change changing to using `flags.NewURLsValue`
the endpoings get here will be int the formort:  `unix://localhost:4085651753091690150`
https://github.com/pingcap/tidb-binlog/blob/649ec60f9ceb65fbbaf22e06e43e39ae62e56df9/pump/node_test.go#L58

there's some similar code it wil use by `flag.NewUrLsValue`...
https://github.com/pingcap/tidb-binlog/blob/649ec60f9ceb65fbbaf22e06e43e39ae62e56df9/pkg/types/urls.go#L40

july2993:
close for inactive

Issue #891, opened 2020-02-07
TLS certificates online reload


DanielZhangQD:
## Feature Request

**Is your feature request related to a problem? Please describe:**

Usually, there will be an expiration of the TLS certificates, we have to support online reload with new certificates instead of just restart the process.
Suggest all the PingCAP components and tools to implement the cert reload feature in the same way.
**Describe the feature you'd like:**

**Describe alternatives you've considered:**

**Teachability, Documentation, Adoption, Migration Strategy:**

july2993:
blocked by #904,  we support online reload after fixing the current TLS support.

PR #926, opened 2020-03-05
Reload cert/key of tls for new connection in pd/kv client

july2993:
fix https://github.com/pingcap/tidb-binlog/issues/891
Reload cert/key of tls for new connection

follow https://github.com/pingcap/tidb-binlog/pull/927, this pr make pd/kv client support online reload too.

ref https://github.com/pingcap/tidb/issues/14666
update tidb/pd dependency to the newest version that support online reload of cert/key

### Check List 

Tests 

 check the version of the dep support online reload manually.

Related changes


WangXiangUSTC:
LGTM

july2993:
> LGTM, but I don't see how this change (upgrade go.mod dependencies) has any relationship with the PR title.

we can make sure the version after this https://github.com/pingcap/pd/pull/2162 can support reload new TLS certificates.

we just pass the file path into the client, the reload is supported by the client internally.
https://github.com/pingcap/pd/blob/0f3272c9b0ef661bdaea7f4ed66b13955c503b72/client/client.go#L129

Issue #900, opened 2020-02-14
Drainer should support kafka clientID property


dixingxing0:
## Feature Request

**Describe the feature you'd like:**
Our kafka cluster is authenticate by kafka clientID,  since drainer do not support clientID property, we can't push binlog data to kafka cluster.
So drainer should support kafka clientID property in our scenario.

WangXiangUSTC:
We can add a new configuration for Drainer, just like the `kafka-version` https://github.com/pingcap/tidb-binlog/blob/1239134f5437dada58de6f73025ce3fe646dc397/drainer/sync/util.go#L41

dijingran:
Thanks, I am working on this now, if test going well, i will submit a PR.

WangXiangUSTC:
support in https://github.com/pingcap/tidb-binlog/pull/902

PR #929, opened 2020-03-06
drainer: add KafkaClientID in DBConfig (#902)

WangXiangUSTC:
Improve configuration for kafka syncer:
Add kafka-client-id in drainer.toml to config kafka client.id property.

cherry-pick from #902 
Drainer should support kafka clientID property #900


michaeljinxq:
LGTM

Issue #930, opened 2020-03-09
Validate CommonName attribute for TLS certificate


tennix:
## Feature Request

**Is your feature request related to a problem? Please describe:**

Currently, any TLS certificates issued by the same CA can access TiDB cluster component. This is insecure as the same CA issued certificates are pretty common.

**Describe the feature you'd like:**

Allow pump and drainer add TLS certificate CN validation, this is the issue in PD, TiKV and TiDB pingcap/pd#2209 tikv/tikv#6982 pingcap/tidb#15137

**Describe alternatives you've considered:**

**Teachability, Documentation, Adoption, Migration Strategy:**

shenli:
We need this feature in 3.1.0.

PR #934, opened 2020-03-17
support common name check for tls

WangXiangUSTC:
fix issue https://github.com/pingcap/tidb-binlog/issues/930
Allow pump and drainer add TLS certificate CN validation

add `VerifyPeerCertificate` function for tls

### Check List 

Tests 

 - Integration test

Related changes

### DIFF
--- a/cmd/drainer/drainer.toml
+++ b/cmd/drainer/drainer.toml
@@ -118,6 +118,8 @@ port = 3306
 # ssl-cert = "/path/to/drainer.pem"
 # Path of file that contains X509 key in PEM format.
 # ssl-key = "/path/to/drainer-key.pem"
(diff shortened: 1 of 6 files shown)


WangXiangUSTC:
@july2993 @kennytm PTAL

WangXiangUSTC:
> pls update config template
> rest LGTM

update in f961dd9

WangXiangUSTC:
> LGTM.
> 
> When are we going to move to tidb-tool's TLS package?

seems that `binlog` use a different way to support online reload, so I need to modified a lot of code if I use pkg in `tidb-tools`, so just update the `pkg/security` in this pr 😃 @kennytm

PR #946, opened 2020-04-06
support syncer plugin

tsthght:
Drainer support plugin to make different syncers can be supported

### Check List 

Tests 

 - Unit test


elvizlai:
PTAL

For some reason, we need send some of decoded message to `ELK`
Now we had to decode binlog file using `reparo` print and resolve printed message.

PR #950, opened 2020-04-08
Handle create/drop sequence job type

july2993:
handle create/drop sequence job type
filter out the special DDL binlog for replicate the value of sequence now.

before this PR: will abort the replicate once create sequence at upstream.
after this PR: the create/drop sequence dll can be replicated, but the value of the sequence WILL NOT be replicated to downstream.

1. handle create/drop sequence job type
2. ref https://github.com/pingcap/tidb/pull/14954
TiDB will write a fake ddl binlog like: select setval(`seq`.`sequence_name`, 1000)
when the value of sequence is changed for replicate the value of the sequence.
we CAN NOT query the job from the tikv according the job id.
just skip this kind of binlog now.

### Check List 

Tests 

 - Integration test

Code changes

Side effects

Related changes


july2993:
@GregoryIan PTAL

GregoryIan:
This implementation of the DDL really irks me, completely out of spec, a DDL, but donn't have DDL job. This implementation forces the binlog or CDC to have special dispersed logic to filter this DDL @siddontang

PR #960, opened 2020-05-06
.*: update max message size for grpc

WangXiangUSTC:

tidb support big transaction, and the binlog's size may bigger than 1G, so need to update the grpc's max message size.

update the max message size to `math.Int32`(2G)

### Check List 

Tests 

 - Unit test

Related changes


WangXiangUSTC:
@july2993 @jackysp PTAL

PR #971, opened 2020-05-29
Fix notify drainer not enable tls

july2993:
Fix notify drainer not enable tls

Fix notify drainer not enable tls

### Check List 

Tests 

 - Integration test
check the new started pump in /tests/restart/run.sh do can be started.
### Release note
- Fix pump notify drainer not enable TLS.

### DIFF
--- a/pump/node.go
+++ b/pump/node.go
@@ -14,6 +14,7 @@
 package pump
 
 import (
+	"crypto/tls"
 	"fmt"
 	"io/ioutil"
 	"net"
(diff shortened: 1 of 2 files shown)


july2993:
["Error 8216: Invalid auto random: auto_random option must be defined on `bigint` column, but not on `int` column"]

july2993:
😳
```
[2020-05-29T04:00:51.258Z] [2020/05/29 11:58:09.889 +08:00] [ERROR] [executor.go:111] ["Exec fail, will rollback"] [query="INSERT INTO `test`.`t`(`a`,`b`) VALUES(?,?)"] [args="[6629298651489370113,\"11\"]"] [error="Error 8216: Invalid auto random: Explicit insertion on auto_random column is disabled. Try to set @@allow_auto_random_explicit_insert = true."]


```

PR #972, opened 2020-05-29
Fix notify drainer not enable tls

july2993:
pick #971 Fix notify drainer not enable tls

Fix notify drainer not enable tls

### Check List 

Tests 

 - Integration test
check the new started pump in /tests/restart/run.sh do can be started.
### Release note
- Fix pump notify drainer not enable TLS.

### DIFF
--- a/pump/node.go
+++ b/pump/node.go
@@ -14,6 +14,7 @@
 package pump
 
 import (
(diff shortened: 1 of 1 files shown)


WangXiangUSTC:
LGTM

PR #974, opened 2020-05-30
Drainer support plugin framework

tsthght:
Drainer support plugin framework.

Users can customize the drainer plugin.

### Check List 

Tests 

 - Unit test


drainer/checkpoint/checkpoint.go
```diff
@@ -50,7 +50,7 @@ func NewCheckPoint(cfg *Config) (CheckPoint, error) {
 		err error
 	)
 	switch cfg.CheckpointType {
-	case "mysql", "tidb":
+	case "mysql", "tidb", "plugin":
```
WangXiangUSTC, line 53: this means we must save checkpoint in a database for plugin type, how about adding a judgment for plugin: when `cfg.DB != nil` use `newMysql`, otherwise use NewFile

drainer/syncplg/plugin.go
```diff
@@ -0,0 +1,72 @@
...
+	queryHistogramVec *prometheus.HistogramVec,
+	sqlMode *string,
+	destDBType string,
+	relayer relay.Relayer,
+	info *loopbacksync.LoopBackSync,
+) (sync.Syncer, error)
+
+//LoadPlugin load syncer plugin
+func LoadPlugin(path, name string) (NewSyncerFunc, error) {
+	fp := path + "/" + name
```
WangXiangUSTC, line 52: how about using `path.Join`

WangXiangUSTC:
@july2993 PTAL

PR #976, opened 2020-06-01
Enable allow_auto_random_explicit_insert if it's suppported

july2993:
pick #975
After https://github.com/pingcap/tidb/pull/17102 unable to replicate auto random table.

the pr add a session system var `allow_auto_random_explicit_insert`, default is false, must enable for insert value explicit, or can't
replicate. This is unfriendly for replication or restores tools. 

the test is disabled in #971 temporarily.

Enable `allow_auto_random_explicit_insert` if ths sys var is supported. 

### Check List 

Tests 

 - Integration test
test with version support and not support this var.

### Release note
- Enable `allow_auto_random_explicit_insert` automatic.

### DIFF
--- a/pkg/loader/util.go
+++ b/pkg/loader/util.go
@@ -25,6 +25,7 @@ import (
 
 	"github.com/go-sql-driver/mysql"
 	"github.com/pingcap/errors"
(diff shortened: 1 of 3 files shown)


leoppro:
LGTM

PR #977, opened 2020-06-05
modify plugin

tsthght:
Drainer support plugin framework.

Users can customize the drainer plugin.

### Check List 

Tests 

 - Unit test

### DIFF
--- a/drainer/sync/plugin_demo.go
+++ b/drainer/sync/plugin_demo.go
@@ -1,0 +1,56 @@
+package sync
+
+import (
+	"fmt"
+
+	"github.com/pingcap/log"
+	"github.com/pingcap/tidb-binlog/drainer/loopbacksync"
+	"github.com/pingcap/tidb-binlog/drainer/relay"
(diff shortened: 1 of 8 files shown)


drainer/sync/plugin_demo.go
```diff
@@ -0,0 +1,56 @@
...
+	return nil
+}
+
+//SetSafeMode is the method that interface must implement
+func (sd *SyncerDemo) SetSafeMode(mode bool) bool {
+	return false
+}
+
+//NewSyncerDemo is a syncer demo
+func NewSyncerDemo(
```
WangXiangUSTC, line 38: how about adding some comments for how to use these arguments?  For example, I don't know what the `file` used for.

tsthght: Thanks, I had writen a READ.md. PTAL @WangXiangUSTC

WangXiangUSTC:
but need to reset the last two commit and then commit with your account 😢

PR #980, opened 2020-06-16
expose merge mode by config and add metrics

july2993:
test by run at upstream:
```
sysbench  oltp_insert --mysql-host=127.0.0.1 --mysql-user=root --mysql-password= --mysql-port=4000 --tables=10 --table-size=1000000 --threads=64 --create_secondary=off --report-interval=10 prepare
```
and start cluster locally and a MySQL as downstream
drainer config:
```
worker-count = 256
txn-batch = 200
```

master:
Screen Shot 2020-06-16 at 11 49 23 PM

dml number per worker is only 10~20, the load of MySQL is low. about 100% cpu

after inc the limit in-memory dml:
Screen Shot 2020-06-17 at 12 46 05 AM
mysql about 600% cpu

Screen Shot 2020-06-17 at 12 49 05 AM

using merge = true:
```
[syncer.to]
merge = false
```
Screen Shot 2020-06-16 at 10 59 47 PM

### Release note
- No Release note.

### DIFF
--- a/pkg/loader/load.go
+++ b/pkg/loader/load.go
@@ -488,7 +488,7 @@ func (s *loaderImpl) singleExec(executor *executor, dmls []*DML) error {
 		byHash[idx] = append(byHash[idx], dml)
 	}
 
-	if s.metrics.QueueSizeGauge != nil {
+	if s.metrics != nil && s.metrics.QueueSizeGauge != nil {
 		// limit 10 sample
 		for i := 0; i < len(byHash) && i < 10; i++ {
 			name := "worker_" + strconv.Itoa(i)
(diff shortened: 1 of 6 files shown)


pkg/loader/load.go
```diff
@@ -477,6 +488,14 @@ func (s *loaderImpl) singleExec(executor *executor, dmls []*DML) error {
 		byHash[idx] = append(byHash[idx], dml)
 	}
 
+	if s.metrics.QueueSizeGauge != nil {
```
GregoryIan, line 491:
```suggestion
	if s.metrics != nil && s.metrics.QueueSizeGauge != nil {
```

Issue #994, opened 2020-08-07
pump supports setting the binlog file GC time according to a smaller time granularity


IANTHEREAL:
## Feature Request

**Is your feature request related to a problem? Please describe:**

When I'm importing data, I find that binlog file size of pump increases rapidly, resulting in insufficient space error.  I want to quickly release space, but now pump only supports setting the GC interval according to the granularity of `day`

**Describe the feature you'd like:**

Supports more fine-grained GC time control. It would be better to be same with TiDB, which can be set in minutes or hours

PR #996, opened 2020-08-07
support go time package format gc time

lichunzhu:
resolve https://github.com/pingcap/tidb-binlog/issues/994

Support gc time for "8m", "7h". If no unit is specified, gc duration will use day as unit as before. 

It should be noticed that the max time duration unit that supported in golang is hour "h". 
https://github.com/golang/go/issues/11473
https://github.com/golang/go/issues/17767

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

Related changes

### Release note
- support go time duration format for pump GC configuration


pkg/util/util.go
```diff
@@ -301,3 +302,16 @@ func WaitUntilTimeout(name string, fn func(), timeout time.Duration) {
...
+
+// ParseGCDuration parses gc durations. The default unit is day.
+func ParseGCDuration(gc string) (time.Duration, error) {
+	d, err := strconv.ParseUint(gc, 10, 64)
+	if err == nil {
+		return time.Duration(d) * 24 * time.Hour, nil
+	}
+	gcDuration, err := time.ParseDuration(gc)
+	if err != nil {
+		return 0, errors.Errorf("unsupported gc time %s, please use 7 or 8h(max unit is hour) format gc time. err: %s", gc, err)
```
july2993, line 314: errors.Annotatef(err, "unsupported gc time %s, etc: use 7 for 7day,  7h for 7 hour")

pkg/util/util_test.go
```diff
@@ -110,6 +110,24 @@ func (s *utilSuite) TestStdLogger(c *C) {
...
+
+	gc = "30m"
+	expectDuration = 30 * time.Minute
+	duration, err = ParseGCDuration(gc)
+	c.Assert(err, IsNil)
+	c.Assert(duration, Equals, expectDuration)
+
+	gc = "7d"
+	_, err = ParseGCDuration(gc)
+	c.Assert(err, ErrorMatches, `unsupported gc time 7d, please use 7 or 8h\(max unit is hour\) format gc time.*`)
```
july2993, line 128: just check not nil is enough

pump/config.go
```diff
@@ -65,8 +65,8 @@ type Config struct {
 	EtcdDialTimeout   time.Duration
 	DataDir           string `toml:"data-dir" json:"data-dir"`
 	HeartbeatInterval int    `toml:"heartbeat-interval" json:"heartbeat-interval"`
-	// pump only stores binlog events whose ts >= current time - GC(day)
-	GC       int             `toml:"gc" json:"gc"`
```
july2993, line 69: I think changing the type is a breaking change for toml marshaling and unmarshaling?
but ok for the command line.

maybe you can do this for compatibility:
```
type GC time.Duration
... impl https://pkg.go.dev/encoding?tab=doc#TextUnmarshaler interface for `GC` type
for handling both types
``

or you may keep GC int and add another config item for compatibility

Comment (lichunzhu) [reply pump/config.go:69]: Yes, it is a breaking change for toml unmarshaling... I prefer the first solution. If we add another field I think it will be more confusing for the users.

Comment (july2993) [reply pump/config.go:69]: Prefer first too, you can try it.

**Comment by lichunzhu** [comment]
@july2993 Already support both int and string and add some tests. PTAL again.

File: pump/config.go [hunk 4]
```diff
@@ -176,6 +177,15 @@ func (cfg *Config) Parse(arguments []string) error {
 	util.AdjustString(&cfg.DataDir, defaultDataDir)
 	util.AdjustInt(&cfg.HeartbeatInterval, defaultHeartbeatInterval)
 
+	if cfg.GC.Duration == 0 && cfg.GCStr == "" {
```
july2993, line 180: this park looks confusing, when we both set int the config file and the command, the command line value should take effect instead of the one in the config file.

can we change
``` Duration as
type Duration string
``
and using the same object for both the config file and command line flag?

Comment (lichunzhu) [reply pump/config.go:180]: Actually this part can make it to prioritize the config in the command. If we set type Duration to string, how do we get the actual duration? Unmarshall it every time?

Comment (lichunzhu) [reply pump/config.go:180]: Can you explain it more clearly?

File: pump/config.go [hunk 5]
```diff
@@ -100,7 +101,7 @@ func NewConfig() *Config {
 	fs.StringVar(&cfg.EtcdURLs, "pd-urls", defaultEtcdURLs, "a comma separated list of the PD endpoints")
 	fs.StringVar(&cfg.DataDir, "data-dir", "", "the path to store binlog data")
 	fs.IntVar(&cfg.HeartbeatInterval, "heartbeat-interval", defaultHeartbeatInterval, "number of seconds between heartbeat ticks")
-	fs.IntVar(&cfg.GC, "gc", defaultGC, "recycle binlog files older than gc days")
+	fs.StringVar(&cfg.GCStr, "gc", "", "recycle binlog files older than gc time. default unit is day. also accept 8h format time(max unit is hour)")
```
july2993, line 104: set default value, so user can see the default value like before?

pump/config.go
july2993: > Actually this part can make it to prioritize the config in the command

yes, I got it wrong.

> If we set type Duration to string, how do we get the actual duration? Unmarshall it every time?

I think parsing it every time is ok here.


I am ok with this now, this will not block this pr.

pump/config.go
lichunzhu: If we set default value here, the value in toml will always be covered by default "7". Can we add it in the prompt?

july2993: ok, that's using two objects making it complex。

pkg/util/duration.go
```diff
@@ -0,0 +1,76 @@
...
+// UnmarshalJSON parses a JSON string into the duration.
+func (d *Duration) UnmarshalJSON(text []byte) error {
+	s, err := strconv.Unquote(string(text))
+	if err != nil {
+		return errors.WithStack(err)
+	}
+	*d = Duration(s)
+	return nil
+}
+
```
july2993, line 46: var _ toml.** = &Duration{}

--
add this make sure it impl the interface and let's reader know this.

PR #998, opened 2020-08-10
support go time package format gc time (#996)

lichunzhu:
cherry-pick #996 to release-4.0

---

resolve https://github.com/pingcap/tidb-binlog/issues/994

Support gc time for "8m", "7h". If no unit is specified, gc duration will use day as unit as before. 

It should be noticed that the max time duration unit that supported in golang is hour "h". 
https://github.com/golang/go/issues/11473
https://github.com/golang/go/issues/17767

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

Related changes

### Release note
- support go time duration format for pump GC configuration


july2993:
for 4.0 I will merge master directly for no diverse

PR #1030, opened 2020-12-29
drainer: reduce memory usage when restart

### LINKED ISSUE #978: Please help  reduce use system memory when drainer startup
## Feature Request

**Is your feature request related to a problem? Please describe:**

When drainer startup it will take much memory , if tidb have a lot of history ddl 

**Describe the feature you'd like:**

Please reduce use system memory  when drainer startup.

**Describe alternatives you've considered:**

1. When drainer startup , why need use history ddl ?
2. History ddl just some records in table , why it take so much memory?

https://asktug.com/t/topic/34955

https://asktug.com/t/topic/34741

you06:

close #978

When drainer restart, it'll load all DDL history from TiDB cluster, this will cause a large memory usage.

This PR change the DDL history load strategy, load only recently DDLs into drainer, for the stale DDLs, load from a snapshot schema.

### Check List 

Tests 

### Release note

- Reduce memory usage when drainer restart


you06:
This is a draft implementation.

Because the relationship between `job.BinlogInfo.SchemaVersion` and `job.ID` is unknowable, drainer still to load all DDL history to check whether it's schema version is newer than current binlog, the most work we do makes non sense, and it'll cost a long time loading history.

PR #1039, opened 2021-04-16
drainer: fix kafka message limit problem

lichunzhu:
When we use kafka as downstream, the transaction size exceed 1G can't to transport to downstream.

1. Set grpc message limit to IntMax instead of `max.Int32`.
2. Add configuration `max-message-size` in drainer's config.

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

Related changes

### Release note

- Set grpc message limit to IntMax instead of `max.Int32`. Add configuration `max-message-size` in drainer's config.


drainer/config.go
```diff
@@ -439,8 +439,11 @@ func (cfg *Config) adjustConfig() error {
 	}
 
 	if cfg.SyncerCfg.DestDBType == "kafka" {
-		maxMsgSize = maxKafkaMsgSize
-
+		if cfg.SyncerCfg.To.KafkaMaxMessageSize > 0 {
+			maxMsgSize = cfg.SyncerCfg.To.KafkaMaxMessageSize
```
glorv, line 443: What if `KafkaMaxMessageSize` > MaxInt32 。 I think we need to do this check here

glorv (approved):

cmd/drainer/drainer.toml
```diff
@@ -164,6 +164,7 @@ port = 3306
 # kafka-addrs = "127.0.0.1:9092"
 # kafka-version = "0.8.2.0"
 # kafka-max-messages = 1024
+# kafkf-max-message-size = 1073741824 # configure max kafka **client** message size
```
amyangfei, line 167:
```suggestion
# kafka-max-message-size = 1073741824 # configure max kafka **client** message size
```

King-Dylan:
Maybe we also need cherry pick in 4.0

niubell:
close https://github.com/pingcap/tidb/issues/28659

PR #1040, opened 2021-04-19
drainer: support replica clustered index table

lysu:

ref https://github.com/pingcap/tidb/issues/24133

support receive clustered index's binlog

- handle common handle pk for insert type binlog
- upgrade tidb/pd dep to newest version

### Check List 

Tests 

 - Integration test

Code changes

 -  n/a

Side effects

 - n/a

Related changes

### Release note 

- Support replica clustered index table


drainer/translator/translator.go
```diff
@@ -74,7 +96,18 @@ func insertRowToDatums(table *model.TableInfo, row []byte) (pk types.Datum, datu
 			//
 			// Will get -1 here, note: uint64(int64(-1)) = 18446744073709551615
 			// so we change it to uint64 if the column type is unsigned
-			datums[col.ID] = fixType(pk, col)
+			datums[col.ID] = fixType(pk[0], col)
+		case clusteredPk:
+			for idxColOrdinal, idxCol := range commonPKInfo.Columns {
```
lance6716, line 101: will there be a case that `table.IsCommonHandle` is true but no PK? so `commonPKInfo` is nil

drainer/translator/translator.go
```diff
@@ -74,7 +96,18 @@ func insertRowToDatums(table *model.TableInfo, row []byte) (pk types.Datum, datu
...
+				if idxCol.Length != types.UnspecifiedLength {
+					// primary key's prefixed column already in row data.
+					continue
+				}
+				if tblColOrdinal == idxCol.Offset {
```
lance6716, line 106: can we directly access `table.Columns[idxCol.Offset]`, to eliminate the two-level for loop? seems we only need to check `clusteredPK` per table.

drainer/translator/translator.go
```diff
@@ -44,28 +44,50 @@ func getParser() (p *parser.Parser) {
...
+	if table.IsCommonHandle {
+		for _, idx := range table.Indices {
+			if idx.Primary {
+				commonPKInfo = idx
+				break
+			}
+		}
+		if commonPKInfo != nil {
+			pkLen = len(commonPKInfo.Columns)
+		}
```
lichunzhu, line 63: What if `commonPKInfo` is `nil`? Although it seems impossible but we can still put a check here?

lichunzhu (commented):
Rest LGTM

drainer/translator/translator.go
lysu: yes~it's impossible that a clustered index table without pk,  add a check and return an error at https://github.com/pingcap/tidb-binlog/pull/1040/files#diff-d362ef1721274a90486b677b50c57cfec86fac0f7fcf567587b452fe7611898dR61 to avoid someone do it in future - -

drainer/translator/translator.go
lysu: no~ clustered index must be defined with a PK now, add a protection check at https://github.com/pingcap/tidb-binlog/pull/1040/files#diff-d362ef1721274a90486b677b50c57cfec86fac0f7fcf567587b452fe7611898dR61

lysu:
PTAL @lance6716 @lichunzhu

lance6716 (approved):
ptal @july2993

drainer/translator/translator.go
```diff
@@ -44,37 +44,71 @@ func getParser() (p *parser.Parser) {
...
-	// decode the pk value
-	var remain []byte
-	remain, pk, err = codec.DecodeOne(row)
-	if err != nil {
-		return types.Datum{}, nil, errors.Trace(err)
+	var (
+		commonPKInfo *model.IndexInfo
+		pkLen        = 1
+	)
+	if table.IsCommonHandle {
```
july2993, line 54: is the feature is enable and there's no pk for the table, this is false?

tests/run.sh
```diff
@@ -60,6 +60,8 @@ stop_services() {
 
 start_upstream_tidb() {
     cat - > "$OUT_DIR/tidb-config.toml" <<EOF
+new_collations_enabled_on_first_bootstrap = true
```
july2993, line 63: it should still work if new_collations is disabled right? although now we run all test enabling it

july2993 (approved):

july2993 (approved):
just a little concern about if something is broken when clustered index is not enabled since we now run all tests with clustered index enable, but this config of tidb is still disabled default.

rest LGTM

drainer/translator/translator.go
lysu: yes,  no pk table always be non-clustered index table now.

but maybe it should have a chance for someone to implement a new feature like "drop PK in clustered index table" or others(related to his impl) in the future, and if they produce a table like that... they will meet `errors.New("Unsupported clustered index without primary key")` error and fix that logic

tests/run.sh
lysu: yes, it works for old collations, but new_collations code path seems to be more complex and more valuable to test(it's enabled by default in tidb-operator or tidb-cloud)

july2993 (dismissed):

lysu:
> just a little concern about if something is broken when clustered index is not enabled since we now run all tests with clustered index enable, but this config of tidb is still disabled default.
> 
> rest LGTM

agree with this, current we only enable clustered in new added case

https://github.com/pingcap/tidb-binlog/pull/1040/files#diff-36df5a4a60e9a0c12e0794a4da3a446af7f03a9ab10e7b7ba20e02102a7d47eeR266-R271

and run pk test case with both clustered + nonclustered

https://github.com/pingcap/tidb-binlog/pull/1040/files#diff-36df5a4a60e9a0c12e0794a4da3a446af7f03a9ab10e7b7ba20e02102a7d47eeR683

we didn't delete previous nonclustered index test case, so we can find case fail if nonclustered index logic changed too

lysu:
@july2993 @lance6716 tidb's PR has merged and binlog's tidb dep also be updated, PTAL~ thx

lance6716:
what's the change of tidb side 🤣

/lgtm

lysu:
> what's the change of tidb side 🤣
> 
> /lgtm

https://github.com/pingcap/tidb/pull/24135, tidb just turn off switch which added in 5.0 to disable binlog produce, and tiny fix some logic like "do not need to produce _tidb_rowid column for clustered index table" 😄

july2993 (approved):

PR #1047, opened 2021-05-26
drainer: fix collation clustered pk as wrong value

lysu:

close https://github.com/pingcap/tidb/issues/24890

for new collation, pk column will be sortedKey(suitable for CI compare rule), row-column will restore raw data

use row's column instead of pk's column if row's column exists

### Check List 

Tests 

 - Unit test

Code changes

 - n/a
 
Side effects

 -  n/a

Related changes

 - n/a

### Release note 

- Fix clustered index table replica wrong value for collation PK

### DIFF
--- a/drainer/translator/translator.go
+++ b/drainer/translator/translator.go
@@ -101,6 +101,11 @@ func insertRowToDatums(table *model.TableInfo, row []byte) (datums map[int64]typ
 				continue
 			}
 			tblIdxCol := table.Columns[idxCol.Offset]
+			if _, exists := datums[tblIdxCol.ID]; exists {
+				// use row column instead of pk column if row column exists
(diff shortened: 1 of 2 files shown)


july2993 (approved):

cfzjywxk:
@lysu Do we have other combinations that the `CommonHandle` is just sorting key or the `CommonHandle` itsself does not contain complete value?

lysu:
> @lysu Do we have other combinations that the `CommonHandle` is just sorting key or the `CommonHandle` itsself does not contain complete value?

yes, https://github.com/pingcap/automated-tests/pull/678 (WIP for more test)

and binlog-CI test also can find this problem, but due to https://github.com/pingcap/tidb/issues/24796 it be hidden in previous(it use sessionVars...-0-

lichunzhu (approved):

PR #1058, opened 2021-07-06
An example cmd error

chenmin1992:
URL is not equal to instruction text.

change the URL and make it correct.

### Check List 

Tests 

 - No code

Code changes

Side effects

Related changes


chenmin1992:
Is necessary to have a release note for no code changing?

PR #1086, opened 2021-11-15
syncer: add table router and binlog event filters in drainer

lichunzhu:
To implement sync to oracle drainer need to implement binlog event filters and table router first.

All the configurations are implemented in dm v2 configuration templates.
1. Implement binlog event filters and skip steps in `drainer.syncer`.
2. Implement table router but didn't add it to any `dsyncer` except for oracle dsyncer. After a second thought we think that router configuration has some danger for synchronization. We will open this configuration to other dsyncers after a full GA process.
3. This PR doesn't use `table-migrate-rule` to generate baList. We can take a further discussion in this PR later.

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

Side effects

 - Increased code complexity

Related changes

### Release note

 - No release note


lichunzhu:
/rebuild
will add some tests later

lichunzhu:
@glorv @lance6716 PTAL

drainer/syncer.go
```diff
@@ -550,6 +582,66 @@ func filterTable(pv *pb.PrewriteValue, filter *filter.Filter, schema *Schema) (i
...
+				default:
+					err = errors.Errorf("unknown mutation type: %v", tp)
+					return
+				}
+
+				needSkip, err := skipByFilter(binlogFilter, schemaName, tableName, et, "")
+				if err != nil {
+					return false, errors.Trace(err)
+				}
+				if !needSkip {
```
lance6716, line 617: seems we can

if needSkip {continue}

tests/binlogfilter/drainer.toml
```diff
@@ -0,0 +1,71 @@
...
+
+[syncer.binlog-filter-rule]
+
+[syncer.binlog-filter-rule.truncate-table-filter]
+ignore-event = ["truncate table"]
+ignore-sql = []
+
+[syncer.binlog-filter-rule.add-column-aaa-filter]
+ignore-event = []
+ignore-sql = ["alter table .* add column aaa int"]
```
lance6716, line 55: does TiDB binlog wrote DDL binlog after normalization? (\`aaa\`)

lance6716 (approved):

tests/binlogfilter/drainer.toml
lichunzhu: No. Just the original DDL.

drainer/syncer.go
```diff
@@ -61,6 +66,11 @@ type Syncer struct {
 
 	shutdown chan struct{}
 	closed   chan struct{}
+
+	// TODO: to discuss block allow list implementation in the future
+	// baList       *baf.Filter
```
glorv, line 71: Does this mean we want to use baList to replace the  do-dbs/tables ignore-dbs/tables?

drainer/syncer.go
```diff
@@ -562,6 +654,37 @@ func filterTable(pv *pb.PrewriteValue, filter *filter.Filter, schema *Schema) (i
...
+func skipDDLEvent(sql, schema, table string, p *parser.Parser, binlogFilter *bf.BinlogEvent) (ignore bool, err error) {
+	if binlogFilter == nil {
+		return false, nil
+	}
+	stmt, err := p.ParseOneStmt(sql, "", "")
+	if err != nil {
+		log.L().Error("fail to parse ddl", zap.String("ddl", sql), logutil.ShortError(err))
+		// return error if parse fail and filter fail
+		needSkip, err2 := skipByFilter(binlogFilter, schema, table, bf.NullEvent, sql)
+		return needSkip, err2
```
glorv, line 668: IF needSkip is false here, do we need to return a error as this is an unknown statement.

glorv (commented):
rest LGTM

drainer/syncer.go
lichunzhu: I will delete this part. As we discussed before, we will **NOT** use baList to replace the do-dbs/tables ignore-dbs/tables.

drainer/syncer.go
lichunzhu: I think that doesn't matter. If this can't be executed downstream they will return an error.

glorv (approved):

drainer/syncer.go
```diff
@@ -550,6 +580,66 @@ func filterTable(pv *pb.PrewriteValue, filter *filter.Filter, schema *Schema) (i
...
+			mutation.Sequence = mutation.Sequence[0:filteredIdx]
+			if mutation.InsertedRows != nil {
+				mutation.InsertedRows = mutation.InsertedRows[0:filteredInsertIdx]
+			}
+			if mutation.UpdatedRows != nil {
+				mutation.UpdatedRows = mutation.UpdatedRows[0:filteredUpdateIdx]
+			}
+			if mutation.DeletedRows != nil {
+				mutation.DeletedRows = mutation.DeletedRows[0:filteredDeleteIdx]
+			}
```
glorv, line 640:
```suggestion
			mutation.InsertedRows = mutation.InsertedRows[0:filteredInsertIdx]
			mutation.UpdatedRows = mutation.UpdatedRows[0:filteredUpdateIdx]
			mutation.DeletedRows = mutation.DeletedRows[0:filteredDeleteIdx]
```
If rows is nil, then idx should be 0, so we can still do rows[0:0] safely. 😁

PR #1090, opened 2021-11-18
loader: clear txns and dmls buffer to free memory

### LINKED ISSUE #1087: memory leak in drainer after a big transaction
## Bug Report

Please answer these questions before submitting your issue. Thanks!

1. What did you do?
If possible, provide a recipe for reproducing the error.
upstream execute a big transaction(e.g. delete many rows) daily(e.g. at 12:00 a.m.)

2. What did you expect to see?
drainer memory reduce after the transaction was replicated.

3. What did you see instead?
drainer memory didn't reduce.

4. Please provide the relate downstream type and version of drainer.
(run `drainer -V` in terminal to get drainer's version)
v4.0.14

GMHDBJD:
close #1087 

nil the txns and dmls slice so golang can gc the big txn

Related changes

### Release note

 - fix drainer memory leak after a big transaction

### DIFF
--- a/pkg/loader/load.go
+++ b/pkg/loader/load.go
@@ -739,8 +739,16 @@ func (b *batchManager) execAccumulatedDMLs() (err error) {
 	if b.fDMLsSuccessCallback != nil {
 		b.fDMLsSuccessCallback(b.txns...)
 	}
-	b.txns = nil
-	b.dmls = nil
+
+	// set elements to nil for gc
+	for i := range b.txns {
+		b.txns[i] = nil
+	}
+	for i := range b.dmls {
+		b.dmls[i] = nil
+	}
+	b.txns = b.txns[:0]
+	b.dmls = b.dmls[:0]
 	return nil
 }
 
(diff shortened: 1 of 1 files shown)


pkg/loader/load.go
```diff
@@ -739,8 +739,8 @@ func (b *batchManager) execAccumulatedDMLs() (err error) {
 	if b.fDMLsSuccessCallback != nil {
 		b.fDMLsSuccessCallback(b.txns...)
 	}
-	b.txns = b.txns[:0]
-	b.dmls = b.dmls[:0]
+	b.txns = nil
```
lance6716, line 742: maybe we should assign nil to each element to release reference before `b.txns = b.txns[:0]`

lichunzhu: https://stackoverflow.com/questions/50732829/will-gc-collect-the-objects-while-the-array-set-to-nil-in-golang

lance6716 (approved):

lichunzhu (approved):

lichunzhu:
https://github.com/pingcap/tidb-binlog/blob/9f16f6e/pkg/loader/load.go#L477
Do we need to change this too?

GMHDBJD:
> https://github.com/pingcap/tidb-binlog/blob/9f16f6e/pkg/loader/load.go#L477 Do we need to change this too?

Will byHash be GC after the function return?

PR #1095, opened 2021-12-13
replicate ddl data in binlog to oracle

cartersz:
replicate ddl data in binlog to oracle

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

### DIFF
--- a/drainer/syncer.go
+++ b/drainer/syncer.go
@@ -480,7 +480,7 @@ ForLoop:
 				log.Info("skip ddl by binlog event filter", zap.String("schema", schema), zap.String("table", table),
 					zap.String("sql", sql), zap.Int64("commit ts", commitTS))
 				// A empty sql force it to evict the downstream table info.
-				if s.cfg.DestDBType == "tidb" || s.cfg.DestDBType == "mysql" {
+				if s.cfg.DestDBType == "tidb" || s.cfg.DestDBType == "mysql" || s.cfg.DestDBType == "oracle" {
 					shouldSkip = true
 				} else {
 					appendFakeBinlogIfNeeded(nil, commitTS)
@@ -492,7 +492,7 @@ ForLoop:
 				log.Info("skip ddl by SyncDDL setting to false", zap.String("schema", schema), zap.String("table", table),
 					zap.String("sql", sql), zap.Int64("commit ts", commitTS))
 				// A empty sql force it to evict the downstream table info.
-				if s.cfg.DestDBType == "tidb" || s.cfg.DestDBType == "mysql" {
+				if s.cfg.DestDBType == "tidb" || s.cfg.DestDBType == "mysql" || s.cfg.DestDBType == "oracle" {
 					shouldSkip = true
 				} else {
 					appendFakeBinlogIfNeeded(nil, commitTS)
--- a/drainer/syncer.go
+++ b/drainer/syncer.go
@@ -19,6 +19,8 @@ import (
 	"sync/atomic"
 	"time"
 
+	"github.com/pingcap/tidb/parser/ast"
+
 	"github.com/pingcap/errors"
 	"github.com/pingcap/log"
 	bf "github.com/pingcap/tidb-tools/pkg/binlog-filter"
@@ -472,9 +474,12 @@ ForLoop:
 			// shouldSkip is used specially for database dsyncers like tidb/mysql/oracle
 			// although we skip some ddls, but we still need to update table info
 			// ignore means whether we should should this ddl event after binlogFilter
-			var shouldSkip, ignore bool
+			var (
+				shouldSkip, ignore bool
+				stmt               ast.StmtNode
+			)
 
-			if ignore, err = skipDDLEvent(sql, schema, table, p, s.binlogFilter); err != nil {
+			if stmt, ignore, err = skipDDLEvent(sql, schema, table, p, s.binlogFilter); err != nil {
 				break ForLoop
 			} else if ignore {
 				log.Info("skip ddl by binlog event filter", zap.String("schema", schema), zap.String("table", table),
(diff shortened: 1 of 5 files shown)


pkg/loader/load.go
```diff
@@ -444,6 +450,79 @@ func (s *loaderImpl) execDDL(ddl *DDL) error {
 	return errors.Trace(err)
 }
 
+func (s *loaderImpl) processOracleDDL(ddl *DDL) error {
+	ddlStmt := ddl.SQL
+	newStmt := ""
+	if isTruncateTableStmt(ddlStmt) {
```
lichunzhu, line 456: I don't think we need to parse ddl twice here.

pkg/loader/load.go
```diff
@@ -444,6 +450,79 @@ func (s *loaderImpl) execDDL(ddl *DDL) error {
...
+		newStmt = fmt.Sprintf("BEGIN %s.do_truncate('%s.%s','');END;", ddl.Database, ddl.Database, ddl.Table)
+	} else {
+		ok, astStmt := isTruncateTablePartitionStmt(ddlStmt)
+		if ok {
+			partitions := astStmt.Specs[0].PartitionNames
+			newStmt = fmt.Sprintf("BEGIN %s.do_truncate('%s.%s','%s');END;", ddl.Database, ddl.Database, ddl.Table, partitions[0].O)
```
lichunzhu, line 462: Is that okay if we only alter one partition here?

pkg/loader/load.go
```diff
@@ -444,6 +450,79 @@ func (s *loaderImpl) execDDL(ddl *DDL) error {
...
+			log.Error("DDL exec failed.", zap.String("old sql", ddl.SQL), zap.String("new ddl sql", newStmt), zap.Error(err))
+			if rbErr := tx.Rollback(); rbErr != nil {
+				log.Error("Rollback DDL failed", zap.String("old sql", ddl.SQL), zap.String("new ddl sql", newStmt), zap.Error(rbErr))
+			}
+			return err
+		}
+
+		if err = tx.Commit(); err != nil {
+			return err
+		}
```
lichunzhu, line 485: Don't need a txn here. Besides, why `newStmt := newStmt`?
```suggestion
		tx, err := s.db.Exec(newStmt)
```

pkg/loader/load.go
```diff
@@ -771,7 +852,7 @@ func (b *batchManager) execAccumulatedDMLs() (err error) {
 
 func (b *batchManager) execDDL(txn *Txn) error {
 	if err := b.fExecDDL(txn.DDL); err != nil {
-		if !pkgsql.IgnoreDDLError(err) {
+		if b.destDBType == "oracle" || !pkgsql.IgnoreDDLError(err) {
```
lichunzhu, line 855: Why?

pkg/loader/load.go
cartersz: delete it

pkg/loader/load.go
cartersz: because pkgsql.IgnoreDDLError do not support oracle error. so if it found error, error will be return for oracle

pkg/loader/load.go
cartersz: for txn , it is same as mysql ddl process.

pkg/loader/load.go
```diff
@@ -444,6 +450,35 @@ func (s *loaderImpl) execDDL(ddl *DDL) error {
 	return errors.Trace(err)
 }
 
+func (s *loaderImpl) processOracleDDL(ddl *DDL) error {
+	newStmt := fmt.Sprintf("BEGIN %s.do_truncate('%s.%s','');END;", ddl.Database, ddl.Database, ddl.Table)
```
lichunzhu, line 454: I think we'd better still keep parser here.

pkg/loader/load.go
lichunzhu: don't understand. If `pkgsql.IgnoreDDLError` do not support oracle error, all errors will be returned.

lichunzhu (commented):
Rest LGTM

lichunzhu:
@glorv PTAL

pkg/loader/load_test.go
```diff
@@ -375,6 +375,29 @@ func (s *execDDLSuite) TestShouldExecInTransaction(c *check.C) {
 	c.Assert(err, check.IsNil)
 }
 
+func (s *execDDLSuite) TestOracleTruncateDDL(c *check.C) {
+	db, mock, err := sqlmock.New()
+	c.Assert(err, check.IsNil)
+
+	mock.ExpectBegin()
+	mock.ExpectExec("BEGIN test.do_truncate").WillReturnResult(sqlmock.NewResult(0, 0))
```
lichunzhu, line 383: Why isn't table name includes?

lichunzhu (commented):
Rest LGTM

pkg/loader/load_test.go
cartersz: fix

lichunzhu (approved):

drainer/syncer.go
```diff
@@ -472,27 +474,35 @@ ForLoop:
...
 					zap.String("sql", sql), zap.Int64("commit ts", commitTS))
 				// A empty sql force it to evict the downstream table info.
-				if s.cfg.DestDBType == "tidb" || s.cfg.DestDBType == "mysql" {
+				if s.cfg.DestDBType == "tidb" || s.cfg.DestDBType == "mysql" || s.cfg.DestDBType == "oracle" {
 					shouldSkip = true
 				} else {
 					appendFakeBinlogIfNeeded(nil, commitTS)
 					continue
 				}
+			} else if !ignore && s.cfg.DestDBType == "oracle" {
```
glorv, line 494: I remember that drainer already set binlog filter rule to only allow `truncate` for oracle by default, why still need to do this check? 🤔

glorv (approved):
LGTM

Issue #1129, opened 2022-01-17
Drainer: SyncDDL switcher does not work when syncing to oracle


cartersz:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

1. What did you do?
If possible, provide a recipe for reproducing the error.
set SyncDDL= false in configuration file.
on TiDB client, you operation non "truncate table **" statement, the drainer will stop.

2. What did you expect to see?
the drainer will not stop
3. What did you see instead?
the drainer will stop
4. Please provide the relate downstream type and version of drainer.
(run `drainer -V` in terminal to get drainer's version)

PR #1130, opened 2022-01-17
fix sync-ddl control

cartersz:

Issue Number: ref #1129

add SyncDDL switcher to the process  that is whether drainer server should stop when non 'truncate table **' statement sync to drainer.

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

### Release note

- No release note

### DIFF
--- a/drainer/syncer.go
+++ b/drainer/syncer.go
@@ -456,49 +456,50 @@ ForLoop:
 				continue
 			}
 
+			var (
+				shouldSkip    bool
+				schema, table string
+			)
 			sql := b.job.Query
-			var schema, table string
 			schema, table, err = s.schema.getSchemaTableAndDelete(b.job.BinlogInfo.SchemaVersion)
 			if err != nil {
 				err = errors.Trace(err)
 				break ForLoop
 			}
-
-			if s.filter.SkipSchemaAndTable(schema, table) {
-				log.Info("skip ddl by block allow filter", zap.String("schema", schema), zap.String("table", table),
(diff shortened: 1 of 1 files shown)


lichunzhu (approved):
LGTM

drainer/syncer.go
```diff
@@ -492,7 +492,7 @@ ForLoop:
 					continue
 				}
 			} else if !ignore && s.cfg.DestDBType == "oracle" {
-				if _, ok := stmt.(*ast.TruncateTableStmt); !ok {
+				if _, ok := stmt.(*ast.TruncateTableStmt); !ok && s.cfg.SyncDDL {
```
glorv, line 495: I'm a bit confuse why we only check `s.cfg.SyncDDL` in such inner block. Is it bettern to check and skip all DDL event at L433. @lichunzhu

lichunzhu: Because we don't `continue` here but just set `shouldSkip = true` to make sure this table's schema can be updated in time. Maybe we should return an error in oracle dsyncer's loader, but I think return an error here is also okay.

glorv: I mean we should check `!s.cfg.SyncDDL` to L433 instead of check here.

lichunzhu: SGTM, but we need to sync this item at L433 too.

glorv (approved):

PR #1144, opened 2022-03-17
drainer: filter out placement special comments in ddl

### LINKED ISSUE #1145: Filter out placement special comments when executing ddl in drainer
## Feature Request

When syncing the a tidb downstream, we are using `tidb_placement_mode='ignore'` to skip all placement settings. However, for some tidb version like `5.0` , `tidb_placement_mode` not exists but placement also works. Some error may occurs when the downsteam is `5.0`.

So we should filter placement special comments in sql to avoid this.

lcwangchao:

Issue Number: close #1145

### Check List 

Tests 

 - Unit test

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

### Release note

- No release note


hawkingrei (approved):

lance6716:
can we use parser to remove the comment?

pkg/loader/load.go
```diff
@@ -418,9 +418,18 @@ func (s *loaderImpl) processMysqlDDL(ddl *DDL) error {
...
 
-		if _, err = tx.Exec(ddl.SQL); err != nil {
+		sql, err := removeDDLPlacementOptions(ddl.SQL)
+		if err != nil {
+			return err
+		}
+
+		if sql == "" {
+			log.Info("skip ddl with all placement options", zap.String("sql", ddl.SQL))
+		}
```
lichunzhu, line 428: return `tx.Commit` here?

lichunzhu:
> can we use parser to remove the comment?

+1

lcwangchao:
> can we use parser to remove the comment?

I tried to use parser, but it should first the upgrade mode. However upgrading the depencies casuses more problems and some other ut fails...

pkg/loader/util.go
```diff
@@ -448,3 +449,18 @@ func getOracleUniqKeys(db *gosql.DB, schema, table string) (uniqueKeys []indexIn
...
 }
+
+func removeDDLPlacementOptions(sql string) (string, error) {
+	stmt, err := parser.New().ParseOneStmt(sql, "", "")
+	if err != nil {
+		return "", err
+	}
+
+	var sb strings.Builder
+	restoreCtx := format.NewRestoreCtx(format.DefaultRestoreFlags|format.RestoreTiDBSpecialComment|format.SkipPlacementRuleForRestore, &sb)
```
lichunzhu, line 460: Can we check whether to change DDL by parser checking does this DDL contains `placementRule`?
Maybe most DDLs don't contain placement rules. But after this commit, their DDLs will be all rewritten.

lcwangchao: The simplest way is to check if string contains the string `/*T![placement]`

pkg/loader/load.go
```diff
@@ -418,9 +418,19 @@ func (s *loaderImpl) processMysqlDDL(ddl *DDL) error {
 			}
 		}
 
-		if _, err = tx.Exec(ddl.SQL); err != nil {
+		sql, err := removeDDLPlacementOptions(ddl.SQL)
+		if err != nil {
+			return err
+		}
```
lichunzhu, line 424: Can we just warn this error here and execute the origin SQL? just like we did in
https://github.com/pingcap/tidb-binlog/blob/194d4ac/pkg/util/util.go#L343

lichunzhu (approved):
Rest LGTM

pkg/loader/load.go
lcwangchao: fine

lance6716 (commented):
lgtm, I'll approve after https://github.com/pingcap/tidb-binlog/pull/1144#discussion_r829782348 is resolved

pkg/loader/load.go
lcwangchao: done

lance6716 (approved):

PR #1172, opened 2022-07-05
Add test for DDL "set tiflash mode"

hongyunyan:

Issue Number: ref #1175 https://github.com/pingcap/tidb/issues/35850

This PR want to add an integration test to make sure "alter table xx set tiflash mode xx" works with the tools.

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

### Release note

- No release note


hongyunyan:
cc @breezewish @lichunzhu

pkg/loader/load.go
```diff
@@ -1008,3 +1008,28 @@ func isSetTiFlashReplica(sql string) bool {
 
 	return false
 }
+
+func isSetTiFlashMode(sql string) bool {
```
lichunzhu, line 1012: Can we merge this function into `isSetTiFlashReplica`? They look so similar except for L1029

hongyunyan: Yes, I think we can merge the function.

pkg/loader/load.go
```diff
@@ -452,7 +452,7 @@ func (s *loaderImpl) processMysqlDDL(ddl *DDL) error {
 		return nil
 	})
 
-	if err != nil && isSetTiFlashReplica(ddl.SQL) {
+	if err != nil && isSetTiFlashReplicaOrTiFlashMode(ddl.SQL) {
```
breezewish, line 455: Maybe we should ignore all errors related with TiFlash? Since TiFlash statements are not available in other dbs.

breezewish (approved):

PR #1177, opened 2022-07-09
drainer: fix semicolon in comment causes parse failed

GMHDBJD:

Issue Number: close #xxx

### Check List 

Tests 

 - Unit test
 - Manual test

### Release note

- No release note

### DIFF
--- a/drainer/util.go
+++ b/drainer/util.go
@@ -257,11 +257,11 @@ func getTableFromDumpFilename(filename string) (db, table string, ok bool) {
 	idx := strings.LastIndex(filename, "-schema.sql")
 	name := filename[:idx]
 	// for db.db.table
-	idx = strings.Index(name, ".")
-	if idx == -1 || idx+1 >= len(name) {
+	fields := strings.SplitN(name, ".", 2)
+	if len(fields) != 2 {
 		return "", "", false
 	}
-	return name[:idx], name[idx+1:], true
+	return fields[0], fields[1], true
 }
 
 type schemaKey struct {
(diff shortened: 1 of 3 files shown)


drainer/util.go
```diff
@@ -273,20 +274,37 @@ type schemaInfo struct {
 	id   int64
 }
 
+// copy from dm:loader
 func getStmtFromFile(file string) (string, error) {
-	content, err := ioutil.ReadFile(file)
+	data := make([]byte, 0, 1024*1024)
```
kennytm, line 279: use a `strings.Builder` here

`data = append(data, []byte(realLine)...)` can be replaced by `data.WriteString(realLine)`.

`data[len(data)-1] == ';'` can be replaced by `strings.HasSuffix(realLine, ";")`

`query := string(data)` can be replaced by `query := data.String()`

`data = data[0:0]` can be replaced by `data.Reset()`

drainer/util.go
```diff
@@ -256,11 +256,12 @@ func getTableFromDumpFilename(filename string) (db, table string, ok bool) {
...
 	name := filename[:idx]
-	fields := strings.Split(name, ".")
-	if len(fields) != 2 {
+	// for db.db.table
+	idx = strings.Index(name, ".")
+	if idx == -1 || idx+1 >= len(name) {
 		return "", "", false
 	}
-	return fields[0], fields[1], true
+	return name[:idx], name[idx+1:], true
```
kennytm, line 264:
```go
fields := strings.SplitN(name, ".", 2)
//                     ^          ^^^
if len(fields) != 2 {
    return "", "", false
}
return fields[0], fields[1], true
```

kennytm (approved):

